PR #4, opened 2019-10-03
Feature/require over import

sblaurock:
- `import` and `export` are only natively supported in Node 12. Revert to using `require` to increase compatibility with older versions of Node.

All tests pass. Code coverage is stable.

### DIFF
--- a/src/modules/autocomplete.js
+++ b/src/modules/autocomplete.js
@@ -85,7 +85,7 @@ const autocomplete = (options) => {
             return response.json();
           }
 
-          throw response;
+          throw new Error(response.statusText);
         })
         .then((json) => {
           if (json.sections) {
--- a/src/modules/autocomplete.js
+++ b/src/modules/autocomplete.js
@@ -85,7 +85,7 @@ const autocomplete = (options) => {
             return response.json();
           }
 
-          throw new Error(response.statusText);
+          throw response;
         })
         .then((json) => {
           if (json.sections) {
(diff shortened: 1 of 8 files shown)


src/modules/search.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable import/prefer-default-export, object-curly-newline */
```
rubikzube: we can remove the disabling of `import/prefer-default-export` here

src/modules/autocomplete.js
```diff
@@ -85,7 +85,7 @@ export function autocomplete(options) {
             return response.json();
           }
 
-          throw new Error(response.statusText);
+          throw response;
```
rubikzube: Why are we throwing the entire response here?  it's not an error.

sblaurock: it is - if not, `response.ok` would be true.

src/modules/search.js
sblaurock: Nice catch!

src/modules/recommendations.js
```diff
@@ -64,7 +64,7 @@ export function recommendations(options) {
         return response.json();
       }
 
-      throw new Error(response.statusText);
+      throw response;
```
rubikzube: Why are we throwing the entire response here?

src/modules/search.js
```diff
@@ -161,7 +161,7 @@ export function search(options) {
             return response.json();
           }
 
-          throw new Error(response.statusText);
+          throw response;
```
rubikzube: Why are we throwing the entire response here?

src/modules/search.js
```diff
@@ -201,7 +201,7 @@ export function search(options) {
...
```
rubikzube: Why are we throwing the entire response here?

PR #15, opened 2019-10-18
Feature/recommendation ~placements~ pods

sblaurock:
- Recommendations endpoint changes to utilize pod identifier in favor of declaring specific method.
- `results` becomes `numResults`.

*Do not merge* - blocked until backend `/pods/` endpoint has been deployed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ constructorio.recommendations.getRecommendations('pod-id', { parameters }).then(
 
 | Parameter | Type | Description |
 | --- | --- | --- |
-| `results` | number | Number of results to retrieve |
+| `numResults` | number | Number of results to retrieve |
 | `itemIds` | string or array | Item ID(s) to retrieve recommendations for |
 | `section` | string | Section to display results from |
 
--- a/README.md
+++ b/README.md
@@ -88,38 +88,11 @@ constructorio.autocomplete.getResults('dogs', {
 
 ### Recommendations
 
-The recommendations module can be used to retrieve item recommendation results. Responses will be delivered via a Promise. The `parameters` object is optional. Item id's may be of type string or array.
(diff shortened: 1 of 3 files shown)


spec/src/modules/recommendations.js
```diff
@@ -126,7 +127,10 @@ describe('ConstructorIO - Recommendations', () => {
         fetch: fetchSpy,
       });
 
-      recommendations.getAlternativeItems(itemId, { results }).then((res) => {
+      recommendations.getRecommendations(placement, {
+        itemIds,
+        results,
```
jjl014: saw this and was immediately confused by `results`, had to look at the code to remember that it was `numResults`...

we had a discussion on this before but i don't remember why we didn't go with `numResults` 🤔

rubikzube: Good point.  We should change to `numResults`.

README.md
```diff
@@ -88,38 +88,11 @@ constructorio.autocomplete.getResults('dogs', {
 
 ### Recommendations
 
-The recommendations module can be used to retrieve item recommendation results. Responses will be delivered via a Promise. The `parameters` object is optional. Item id's may be of type string or array.
+The recommendations module can be used to retrieve item recommendation results for a supplied pod identifier. Responses will be delivered via a Promise. The `parameters` object is optional. Item id's may be of type string or array.
```
rubikzube: I would change `item recommendations` here to just be `recommendations`.  Being less descriptive might be better for future compatibility.

rubikzube: Might change `supplied pod identifier` to `given pod`

spec/src/modules/recommendations.js
sblaurock: 👍 I'll open a separate PR to change this for autocomplete as well.

rubikzube:
Looks great overall, let's not merge until tomorrow

README.md
sblaurock: 👍

PR #20, opened 2019-10-30
Add recommendations tracking methods

sblaurock:
- Support POST requests to new v2 endpoints (parameters are contained within body, not URL)
- Request queue items are now objects instead of strings to allow storing of method and body parameters.
- "Legacy" request queue items (strings) are still supported due to backwards compatibility shim.
- Document new endpoints in readme.
- Tests for new endpoints.

Coverage:
98.66% Statements 588/596
86.32% Branches 303/351
98.39% Functions 61/62
98.4% Lines 493/501

Tests:
146/146 passing

### DIFF
--- a/spec/src/modules/tracker.js
+++ b/spec/src/modules/tracker.js
@@ -525,7 +525,7 @@ describe('ConstructorIO - Tracker', () => {
         fetch: fetchSpy,
       });
 
-      expect(tracker.trackConversion(term, parameters)).to.equal(true);
+      expect(tracker.trackConversion(term, {})).to.equal(true);
 
       const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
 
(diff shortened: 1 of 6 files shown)


spec/src/modules/tracker.js
```diff
@@ -525,7 +525,7 @@ describe('ConstructorIO - Tracker', () => {
         fetch: fetchSpy,
       });
 
-      expect(tracker.trackConversion(term, parameters)).to.equal(true);
+      expect(tracker.trackConversion(term, {})).to.equal(true);
```
sblaurock: We were passing `section` as part of the parameters, which defeats the point of the test as we're testing to see that `section` is defaulted when not supplied.

spec/src/modules/tracker.js
```diff
@@ -615,14 +615,14 @@ describe('ConstructorIO - Tracker', () => {
         fetch: fetchSpy,
       });
 
-      expect(tracker.trackPurchase(parameters)).to.equal(true);
+      expect(tracker.trackPurchase({})).to.equal(true);
```
sblaurock: We were passing `section` as part of the parameters, which defeats the point of the test as we're testing to see that `section` is defaulted when not supplied.

PR #22, opened 2019-11-04
Fix/defer queue send

sblaurock:
Tested in sandbox and resolves with race condition issue with beforeunload.

setTimeout 0 (setImmediate) also works, but that feels like it's cutting it a bit close for my taste.

### DIFF
--- a/spec/src/modules/tracker.js
+++ b/spec/src/modules/tracker.js
@@ -20,6 +20,7 @@ const { fetch } = fetchPonyfill({ Promise });
 
 describe('ConstructorIO - Tracker', () => {
   const clientVersion = 'cio-mocha';
+  const waitInterval = 500;
   let fetchSpy;
 
   jsdom({ url: 'http://localhost' });
@@ -40,7 +41,7 @@ describe('ConstructorIO - Tracker', () => {
   });
 
   describe('sendSessionStart', () => {
-    it('Should respond with a valid response', () => {
+    it('Should respond with a valid response', (done) => {
       const { tracker } = new ConstructorIO({
         apiKey: testApiKey,
         fetch: fetchSpy,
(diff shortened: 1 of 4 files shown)


src/utils/request-queue.js
```diff
@@ -35,47 +35,50 @@ class RequestQueue {
...
+            this.requestPending = false;
 
-          this.send();
-        });
+            this.send();
+          });
+        }
       }
-    }
+    }, 200);
```
rubikzube: What are the chances we can make the timeout configurable and have a default?

sblaurock: The chances are 100% =)

sblaurock: Ugh. This is so hard to name.

spec/src/modules/tracker.js
```diff
@@ -20,6 +20,7 @@ const { fetch } = fetchPonyfill({ Promise });
```
sblaurock: The only changes are that we wrap expectations in a `setTimeout` of `waitInterval` to ensure the component has had time to actually make the request (since they are now deferred by a certain amount of time).

src/utils/request-queue.js
```diff
@@ -35,47 +35,50 @@ class RequestQueue {
...
+            this.requestPending = false;
 
-          this.send();
-        });
+            this.send();
+          });
+        }
       }
-    }
+    }, (this.options && this.options.requestQueueDeferTime) || 200);
```
rubikzube: Maybe `requestQueueSendDelay` or `trackingSendDelay`?  Also how did we arrive at `200` as the default?

sblaurock: 👍coming up with a name for that was difficult - I like yours better.

`200` just came out of my head as `0` feels a little short for safety, but we could try `0` for now and see how things go?

PR #23, opened 2019-11-05
Bring in event emitter, emit events on tracker network requests.

sblaurock:
All tests pass (154/154).

Coverage:
98.69% Statements
85.87% Branches
98.55% Functions
98.45% Lines

### DIFF
--- a/spec/src/modules/tracker.js
+++ b/spec/src/modules/tracker.js
@@ -18,7 +18,7 @@ dotenv.config();
 const testApiKey = process.env.TEST_API_KEY;
 const { fetch } = fetchPonyfill({ Promise });
 
-describe('ConstructorIO - Tracker', () => {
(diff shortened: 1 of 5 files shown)


spec/src/modules/tracker.js
```diff
@@ -18,7 +18,7 @@ dotenv.config();
 const testApiKey = process.env.TEST_API_KEY;
 const { fetch } = fetchPonyfill({ Promise });
 
-describe('ConstructorIO - Tracker', () => {
+describe.only('ConstructorIO - Tracker', () => {
```
sblaurock: Shouldn't be committed - will fix before merge.

rubikzube: Looks mostly great.  Might use `emitter` instead of `ee`, tho.  The latter might be too terse for legibility

PR #27, opened 2019-11-13
Document missing methods, remove descriptions.

sblaurock:
`sendSessionStart` and `sendInputFocus` were missing from the documentation.

Question: Should these be renamed to `trackSessionStart` and `trackInputFocus`?

Also removed descriptions from tracking documentation until we can define them.


README.md
```diff
@@ -122,21 +122,31 @@ constructorio.recommendations.getRecommendations('pod-id', { parameters }).then(
 
 The tracker module can be used to send tracking events. Returns `true` when successful, or will throw an error if an issue is encountered.
 
+#### Send session start event
+```javascript
+constructorio.tracker.sendSessionStart();
+```
```
sblaurock: Take out for now.

sblaurock:
> Looks good to me. I think we might also want to add in trackSessionStart as well to the README.

Spoke with Zubin about this and led to a discussion that we should probably be sending this event automatically, but we decided to not document it for now.

PR #29, opened 2019-11-18
Update endpoint, stringify body.

sblaurock:


src/utils/request-queue.js
```diff
@@ -71,7 +71,7 @@ class RequestQueue {
         if (nextInQueue.method === 'POST') {
           request = fetch(nextInQueue.url, {
             method: nextInQueue.method,
-            body: nextInQueue.body,
+            body: JSON.stringify(nextInQueue.body),
```
sblaurock: Re: https://github.com/facebook/react-native/issues/6025

PR #30, opened 2019-11-20
Send universal request keys in query params also for POST

rubikzube:


package.json
```diff
@@ -24,6 +27,57 @@
...
+      "max-params": [
+        "error",
+        4
+      ],
+      "complexity": [
+        "error",
+        20
+      ]
+    }
+  },
```
sblaurock: Any reason it's better to have this in here rather than an rc file?

package.json
```diff
@@ -32,13 +86,13 @@
     "chai": "^4.2.0",
     "chai-as-promised": "^7.1.1",
     "dotenv": "^8.1.0",
-    "eslint": "^6.1.0",
-    "eslint-config-airbnb": "^18.0.1",
-    "eslint-config-airbnb-base": "^14.0.0",
-    "eslint-plugin-import": "^2.18.2",
+    "eslint": "^5.12.1",
```
sblaurock: Down a version?

src/modules/tracker.js
```diff
@@ -454,7 +454,7 @@ class Tracker {
         bodyParams.num_results_viewed = num_results_viewed;
       }
 
-      this.requests.queue(url, 'POST', applyParams(bodyParams, this.options));
+      this.requests.queue(`${url}${applyParamsAsString({}, this.options)}`, 'POST', bodyParams);
```
sblaurock: Doesn't seem right. Inherited properties from constructor aren't being applied to bodyParams since we took applyParams out (`ui`, `key`, `us`, etc.).

Running the tests should reveal more detail.

sblaurock: We're also missing a `?` here. This will generate something like: `https://ac.cnstrc.com/v2/behavioral_action/browse_result_clickc=cio-mocha&i=420334ed-a726-4f3b-ac8f-4d3b49c643c8&s=1&key=ZqXaOfXuBWD4s3XzCI1q&_dt=1574280607096`

package.json
rubikzube: It makes this repo conform to the other repos.

package.json
rubikzube: It makes this repo conform to the other repos.

src/modules/tracker.js
rubikzube: YAAAAAAAS!

package.json
sblaurock: By this logic we update singular dependencies of all repositories at once across the entire org? 😄

sblaurock: (I'm fine with it)

rubikzube: It also doesn't work on my machine as configured, nor Jimmys if I remember a prior conversation correctly.

src/modules/tracker.js
rubikzube: Test don't work, the PR is still a work in progress

package.json
sblaurock: 👍

PR #31, opened 2019-11-22
Improved tracking tests

sblaurock:
- Required and optional parameters for tracking tests are done independently.
- Utilize event emitter to determine test success or error (exposed quite a few issues which leads to a much more "bulletproof" implementation of tracking calls").

Note: `us` parameter isn't supported for a few method calls, hence the skipped tests.

All tests pass (170/170).

Coverage:
98.6% Statements 704/714
83.84% Branches 358/427
98.59% Functions 70/71
98.31% Lines 582/592


src/utils/request-queue.js
```diff
@@ -91,14 +91,16 @@ class RequestQueue {
 
             // Request was successful, but returned a non-2XX status code
             else {
+              const instance = this;
```
rubikzube: Maybe shift this up a little in scope so that we're not using a combination of `instance.eventEmitter` and `this.eventEmitter`?

rubikzube:
We might want to consider splitting up the tracker test file into API v1 and API V2, because it's closing in on 2000 lines

sblaurock:
> We might want to consider splitting up the tracker test file into API v1 and API V2, because it's closing in on 2000 lines

Not a bad call. He's getting quite large and will only get larger if we decide to add tests for specific arguments.

sblaurock:
> > We might want to consider splitting up the tracker test file into API v1 and API V2, because it's closing in on 2000 lines
> 
> Not a bad call. He's getting quite large and will only get larger if we decide to add tests for specific arguments.

@rubikzube Can I do this in a follow-up PR? Will be easier to follow what's going on if we ship this behemoth and then make this change.

spec/src/modules/tracker.js
```diff
@@ -746,75 +1105,107 @@ describe('ConstructorIO - Tracker', () => {
...
+        expect(responseParams).to.have.property('message');
 
-        expect(requestedBodyParams).to.have.property('section').to.equal('Products');
         done();
       }, waitInterval);
     });
 
-    it('Should respond with a valid response when parameters and segments are provided', (done) => {
+    // TODO: Test skipped as segments (us) parameters not supported
+    it.skip('Should respond with a valid response when required parameters and segments are provided', (done) => {
```
jjl014: whoa, that skip is super neat!

PR #39, opened 2020-03-05
Send Tracking Events Option

jjl014:
Updates:
* Added `sendTrackingEvents` option
* Updated tests and added a new one

Pls view diff with `?w=1`

Browse tracking (https://github.com/Constructor-io/autocomplete-ui/pull/470) is blocked by this

### DIFF
--- a/spec/src/utils/request-queue.js
+++ b/spec/src/utils/request-queue.js
@@ -15,6 +15,7 @@ dotenv.config();
 describe('ConstructorIO - Utils - Request Queue', () => {
   const storageKey = '_constructorio_requests';
   const waitInterval = 700;
+  const options = { sendTrackingEvents: true };
 
   describe('queue', () => {
     let defaultAgent;
@@ -37,7 +38,7 @@ describe('ConstructorIO - Utils - Request Queue', () => {
     });
 
     it('Should add url requests to the queue', () => {
-      const requests = new RequestQueue();
+      const requests = new RequestQueue(options);
 
       requests.queue('https://ac.cnstrc.com/behavior?action=session_start');
       requests.queue('https://ac.cnstrc.com/behavior?action=focus');
@@ -49,7 +50,7 @@ describe('ConstructorIO - Utils - Request Queue', () => {
     });
 
     it('Should add object requests to the queue - POST with body', () => {
-      const requests = new RequestQueue();
+      const requests = new RequestQueue(options);
 
       requests.queue('https://ac.cnstrc.com/behavior', 'POST', { action: 'session_start' });
       requests.queue('https://ac.cnstrc.com/behavior', 'POST', { action: 'focus' });
@@ -61,7 +62,7 @@ describe('ConstructorIO - Utils - Request Queue', () => {
     });
 
     it('Should not add requests to the queue if the user has a bot-like useragent', () => {
-      const requests = new RequestQueue();
+      const requests = new RequestQueue(options);
 
       window.navigator.__defineGetter__('userAgent', () => 'Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; Googlebot/2.1; +http://www.google.com/bot.html) Safari/537.36');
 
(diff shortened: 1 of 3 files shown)


src/constructorio.js
```diff
@@ -60,6 +61,7 @@ class ConstructorIO {
       testCells,
       fetch,
       trackingSendDelay,
+      sendTrackingEvents: sendTrackingEvents === false ? false : true, /* eslint-disable-line no-unneeded-ternary */
```
jjl014: Defaulting to `true` unless the option is specifically false

spec/src/utils/request-queue.js
```diff
@@ -37,7 +38,7 @@ describe('ConstructorIO - Utils - Request Queue', () => {
     });
 
     it('Should add url requests to the queue', () => {
-      const requests = new RequestQueue();
+      const requests = new RequestQueue(options);
```
rubikzube: necessary?

src/utils/request-queue.js
```diff
@@ -39,86 +39,88 @@ class RequestQueue {
...
-        }
-
-        if (nextInQueue.method === 'POST') {
-          request = fetch(nextInQueue.url, {
-            method: nextInQueue.method,
-            body: JSON.stringify(nextInQueue.body),
-            headers: { 'Content-Type': 'text/plain' },
-          });
-        }
+    if (this.options.sendTrackingEvents) {
```
rubikzube: no queue for you!

spec/src/utils/request-queue.js
rubikzube: options?

jjl014: yup, since we're instantiating the request queue directly it doesn't have the options that are passed through from the constructor or tracker module

jjl014: same as above

src/utils/request-queue.js
jjl014: moved from `send` to `queue` so if `sendTrackingEvents` is false, it won't queue any events at all

sblaurock:
I don't really see the need to have to pass `options` to all the tests if it's defined as true when not provided?

src/constructorio.js
sblaurock: I think I'd prefer a line above that does

```
if (!sendTrackingEvents) {
    sendTrackingEvents = true;
}
```

above, so we don't have to do the weird unneeded ternary dance?

sblaurock:
I also kind of feel like the intent is backwards here. Perhaps the flag should be `disableSendTrackingEvents` since it's the minority?

spec/src/utils/request-queue.js
sblaurock: It shouldn't need them though since it's defaulted?

sblaurock:
Like, ideally this will always send tracking events unless we explicitly tell it not to (in the one test).

jjl014:
> I don't really see the need to have to pass `options` to all the tests if it's defined as true when not provided?

@sblaurock It's defined as true on the top level constructorio module which doesn't get passed down since the tests , but I can move that logic to the request queue instead if that makes more sense. I was thinking putting it in the constructorio module level might be better since that's where we set all the defaults.

> so we don't have to do the weird unneeded ternary dance?

For the ternary, it's to default to true unless specified to be false. Otherwise values like null or undefined won't be defaulted to true

> I also kind of feel like the intent is backwards here. Perhaps the flag should be disableSendTrackingEvents since it's the minority?

We can go that route as well, I just went with `sendTrackingEvents` to be consistent with the autocomplete-ui

sblaurock:
We can go that route as well, I just went with sendTrackingEvents to be consistent with the autocomplete-ui

> Ah, gotcha. I'm down with this way then. I would like to see a solution that only requires us to provide options for the one test where we don't want this behavior rather than having to pass it to all.

jjl014:
> Ah, gotcha. I'm down with this way then. I would like to see a solution that only requires us to provide options for the one test where we don't want this behavior rather than having to pass it to all.

Yea, that makes sense. Moving setting of the default to true from the constructorio module to the request queue would fix that issue. Lmk if you'd like me to make that change.

sblaurock:
The Jimmy and I have come to an agreement 🚢

PR #43, opened 2020-06-07
Document which tracking parameters are required in README.

sblaurock:

PR #44, opened 2020-06-08
Send Variation IDs with conversions

jjl014:
There is a failing test but it looks like it's handled in #42

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,14 +134,14 @@ constructorio.tracker.trackAutocompleteSelect('dogs', {
 });
 ```
 
-| Parameter | Type |
-| --- | --- |
-| `original_query` | string |
-| `result_id` | string |
-| `section` | string |
-| `tr` | string |
-| `group_id` | string |
-| `display_name` | string |
+| Parameter | Type | Required |
+| --- | --- | :---: |
+| `original_query` | string | ✔ |
+| `result_id` | string | ✔ |
+| `section` | string | ✔ |
+| `tr` | string | |
+| `group_id` | string | |
+| `display_name` | string | |
 
 #### Send autocomplete search event
 ```javascript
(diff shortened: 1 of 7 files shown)


spec/src/modules/tracker.js
```diff
@@ -850,6 +850,7 @@ describe('ConstructorIO - Tracker', () => {
     const requiredParameters = {
       name: 'name',
       customer_id: 'customer-id',
+      variation_id: 'variation-id',
```
sblaurock, line 853: Is `variation_id` a required param? If not, could we add a separate test for optional params (even if it's just this one) for now?

README.md
```diff
@@ -193,6 +193,7 @@ constructorio.tracker.trackConversion('dogs', {
 | --- | --- |
 | `name` | string |
 | `customer_id` | string |
+| `variation_id` | string |
```
sblaurock, line 196: Can we update to play nicely with https://github.com/Constructor-io/constructorio-client-javascript/pull/43?

spec/src/modules/tracker.js
jjl014: sounds good. updated. 👍

README.md
jjl014: Updated, hopefully what i pushed up works fine when everything gets merged into master

README.md
```diff
@@ -193,6 +193,7 @@ constructorio.tracker.trackConversion('dogs', {
 | --- | --- |
 | `name` | string |
 | `customer_id` | string |
+| `variation_id` | string | ✔ |
```
jjl014, line 196: Even with the checkmark there, it still displays as if it were only two columns. So it should work fine when everything is merged into master 👍

sblaurock: Sweet - let me get that merged and then it should make things easier for you over here.

spec/src/modules/tracker.js
```diff
@@ -891,6 +895,43 @@ describe('ConstructorIO - Tracker', () => {
...
+        expect(requestParams).to.have.property('i');
+        expect(requestParams).to.have.property('s');
+        expect(requestParams).to.have.property('c').to.equal(clientVersion);
+        expect(requestParams).to.have.property('_dt');
+        expect(requestParams).to.have.property('name').to.equal(requiredParameters.name);
+        expect(requestParams).to.have.property('customer_id').to.equal(requiredParameters.customer_id);
+        expect(requestParams).to.have.property('result_id').to.equal(requiredParameters.result_id);
+        expect(requestParams).to.have.property('revenue').to.equal(requiredParameters.revenue.toString());
+        expect(requestParams).to.have.property('section').to.equal(requiredParameters.section);
+        expect(requestParams).to.have.property('variation_id').to.equal(optionalParameters.variation_id);
```
sblaurock, line 924: 🔥

PR #45, opened 2020-06-18
Emit events on request response

sblaurock:
Emit custom events on `window` for search, browse, recommendations modules with response data.

3 tests added. All pass.

Lint passes.

PR #46, opened 2020-06-18
Emit events on request response - queued with option to wait for beacon event

sblaurock:
- Events are now queued
- `eventDispatcher` parameters can now be specified on instantiation.
- `eventDispatcher.enabled` defaults to TRUE. If set to FALSE, no events will ever be dispatched
- `eventDispatcher.waitForBeacon` will disable event dispatching until a custom event (`cio.beacon.loaded`) is received on `window`. Once received, all queued events will be dispatched until queue is exhausted.

Full test coverage added for new class (`EventDispatcher`) - 15 in total.

Continued from #45

### DIFF
--- a/src/modules/browse.js
+++ b/src/modules/browse.js
@@ -100,6 +100,7 @@ function createBrowseUrl(filterName, filterValue, parameters, options) {
 class Browse {
   constructor(options) {
     this.options = options;
+    this.module = 'browse';
   }
 
   /**
@@ -119,6 +120,7 @@ class Browse {
    */
   getBrowseResults(filterName, filterValue, parameters) {
     let requestUrl;
+    const method = 'getBrowseResults';
     const fetch = (this.options && this.options.fetch) || fetchPonyfill({ Promise }).fetch;
 
     try {
@@ -144,6 +146,8 @@ class Browse {
               result.result_id = json.result_id;
             });
           }
+
+          helpers.dispatchEvent(this.module, method, 'response', json);
 
           return json;
         }
--- a/src/modules/browse.js
+++ b/src/modules/browse.js
@@ -149,7 +149,7 @@ class Browse {
             });
           }
 
-          this.eventDispatcher.queue(this.module, method, 'response', json);
+          this.eventDispatcher.queue(this.module, method, 'completed', json);
 
           return json;
         }
--- a/src/modules/browse.js
+++ b/src/modules/browse.js
@@ -101,7 +101,6 @@ function createBrowseUrl(filterName, filterValue, parameters, options) {
 class Browse {
   constructor(options) {
     this.options = options;
-    this.module = 'browse';
     this.eventDispatcher = new EventDispatcher(options && options.eventDispatcher);
   }
 
@@ -122,7 +121,6 @@ class Browse {
    */
   getBrowseResults(filterName, filterValue, parameters) {
     let requestUrl;
-    const method = 'getBrowseResults';
     const fetch = (this.options && this.options.fetch) || fetchPonyfill({ Promise }).fetch;
 
     try {
@@ -149,7 +147,7 @@ class Browse {
             });
           }
 
-          this.eventDispatcher.queue(this.module, method, 'completed', json);
+          this.eventDispatcher.queue('browse', 'getBrowseResults', 'completed', json);
 
           return json;
         }
--- a/src/modules/browse.js
+++ b/src/modules/browse.js
@@ -100,8 +100,8 @@ function createBrowseUrl(filterName, filterValue, parameters, options) {
  */
 class Browse {
   constructor(options) {
-    this.options = options;
-    this.eventDispatcher = new EventDispatcher(options && options.eventDispatcher);
+    this.options = options || {};
+    this.eventDispatcher = new EventDispatcher(options.eventDispatcher);
   }
 
   /**
--- a/src/modules/browse.js
+++ b/src/modules/browse.js
@@ -102,7 +102,7 @@ class Browse {
   constructor(options) {
     this.options = options;
     this.module = 'browse';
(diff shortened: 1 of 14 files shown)


src/utils/request-queue.js
```diff
@@ -15,8 +15,9 @@ class RequestQueue {
     this.requestPending = false;
     this.pageUnloading = false;
 
-    // eslint-disable-next-line no-unneeded-ternary
-    this.sendTrackingEvents = (options && options.sendTrackingEvents === false) ? false : true;
+    this.sendTrackingEvents = (options && options.sendTrackingEvents === false)
```
rubikzube, line 18: Would `this.sendTrackingEvents = !!options.sendTrackingEvents` have the same effect?

src/utils/event-dispatcher.js
```diff
@@ -0,0 +1,76 @@
...
+    evt.initCustomEvent(eventName, false, false, detail);
+
+    return evt;
+  }
+};
+
+class EventDispatcher {
+  constructor(options) {
+    this.events = [];
+    this.enabled = (options && options.enabled === false)
```
rubikzube, line 20: Would `this.enabled = !!options.enabled` have the same effect?

src/utils/event-dispatcher.js
```diff
@@ -0,0 +1,76 @@
...
+      ? false
+      : true; // Defaults to 'true'
+    this.waitForBeacon = (options && options.waitForBeacon === false)
```
rubikzube, line 23: Would `this.waitForBeacon = !!options.waitForBeacon` have the same effect?

src/utils/event-dispatcher.js
```diff
@@ -0,0 +1,76 @@
...
+    // - if false, events will never be dispatched
+    // `active` is a variable determining if events will be dispatched
+    // - if `waitForBeacon` is set to true, `active` will be false until beacon event is received
+    this.active = this.enabled;
+
+    // If `waitForBeacon` option is set, only enable event dispatching once event is received from beacon
+    if (this.waitForBeacon) {
+      this.active = false;
+
+      // Mark if page environment is unloading
```
rubikzube, line 37: Comment seems a bit off ...

src/utils/event-dispatcher.js
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    if (this.active) {
+      this.dispatchEvents();
+    }
+  }
+
+  // Dispatch all custom events within queue on `window` of supplied name with data
+  dispatchEvents() {
+    if (this.events.length) {
```
rubikzube, line 64: Not sure if we need the `if` ... the `while` may take care of that implicitly

src/modules/search.js
```diff
@@ -140,11 +144,15 @@ class Search {
             });
           }
 
+          this.eventDispatcher.queue(this.module, method, 'completed', json);
```
rubikzube, line 147: Might be a little too much indirection here with the variables.  It might be easier to read without them, like 
```javascript
this.eventDispatcher.queue('search', 'getSearchResults`, 'completed', json)
```

src/modules/search.js
```diff
@@ -96,6 +97,8 @@ function createSearchUrl(query, parameters, options) {
 class Search {
   constructor(options) {
     this.options = options;
+    this.module = 'search';
+    this.eventDispatcher = new EventDispatcher(options && options.eventDispatcher);
```
rubikzube, line 101: Maybe we could do 
``` javascript
this.options = options || {};
```
so that we can do
```
this.eventDispatcher = new EventDispatcher(options.eventDispatcher);
```

src/utils/request-queue.js
sblaurock: Jimmy and I chatted about this a while when we put this into place. Essentially, it makes it explicit that you must define a boolean. Otherwise you could do things like `sendTrackingEvents: "no"` and it would be truthy.

src/modules/search.js
sblaurock: Agreed - a bit overabstracted.

src/utils/event-dispatcher.js
```diff
@@ -0,0 +1,74 @@
...
+
+    if (this.active) {
+      this.dispatchEvents();
+    }
+  }
+
+  // Dispatch all custom events within queue on `window` of supplied name with data
+  dispatchEvents() {
+    while (this.events.length) {
+      const item = this.events.pop();
```
jjl014, line 65: should we be using `shift()` here instead since it's a queue and not a stack? although i guess maybe we don't care about the order they're sent out...

src/utils/event-dispatcher.js
jjl014: following the other comment i made, this will always be enabled unless otherwise specified. since the EventDispatcher will probably not always be enabled, we might want to go with what zubin suggested and do `!!options.enabled`

src/utils/event-dispatcher.js
jjl014: following the other comment i made, this will always wait for the beacon unless specified to be false, so i think we'd want to keep it how steve currently has it

src/modules/browse.js
```diff
@@ -100,6 +101,7 @@ function createBrowseUrl(filterName, filterValue, parameters, options) {
 class Browse {
   constructor(options) {
     this.options = options;
+    this.eventDispatcher = new EventDispatcher(options && options.eventDispatcher);
```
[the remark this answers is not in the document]
sblaurock: 👍 nice catch.

src/utils/event-dispatcher.js
sblaurock: Great catch - agreed.

PR #50, opened 2020-07-30
[CH 10667] Origin Referrer

jjl014:
#### Updates:
* Send along `origin_referrer` in behavioral requests
* Add new option `sendReferrerWithTrackingEvents` (defaulted to `true`)
* Update tests

#### Notes:
* ~A lot browse/recommendations tests will be failing until backend makes updates~ (origin_referrer is now supported in backend and kibana)
* Blocks https://github.com/Constructor-io/autocomplete-ui/pull/643


spec/src/modules/tracker.js
```diff
@@ -20,13 +20,15 @@ const sendTimeout = 25;
...
-describe('ConstructorIO - Tracker', () => {
+describe.only('ConstructorIO - Tracker', () => {
   const clientVersion = 'cio-mocha';
   const waitInterval = 1000;
   let fetchSpy = null;
   let eventSpy = null;
 
-  jsdom({ url: 'http://localhost' });
+  jsdom({
+    url: 'http://localhost.test/path/name',
```
rubikzube, line 30: Can we add query params here to confirm that they aren't included?

jjl014: Updated w/ query params as well as `sendReferrerWithTrackingEvents` and tests

rubikzube (approved):

src/utils/helpers.js
```diff
@@ -66,6 +66,14 @@ const utils = {
   },
 
   isNil: value => value == null,
+
+  getWindowLocation: () => {
+    if (typeof window !== 'undefined') {
```
sblaurock, line 71: Is this quoted intentionally?

sblaurock (changes requested):
@jjl014 @rubikzube I made a few changes here if that's alright with you

1. Changed test descriptions to be more in-line with existing descriptions (using "not" instead of "NOT", etc.)

2. Moved defaulting of `sendReferrerWithTrackingEvents` into tracker.js rather than in ConstructorIO class as that's the existing pattern for the other tracking options (not opposed to moving this into ConstructorIO, but would rather we do all of them for consistency)

3. Added a test for the default case where `sendReferrerWithTrackingEvents` is not defined at all, to ensure that they are sent by default (only did this on trackSessionStart, not every method).

Testing that these params are being sent on every tracking method feels a bit redundant, but I suppose it's a good thing to make sure every endpoint is willing to accept this parameter.

Now that we have 3 different options related to tracking I would love to explore putting these inside of an object like we have for `eventDispatcher` - some time in the future.

This looks great - just had the one question about `'undefined'` as I believe that might be in error.

src/utils/helpers.js
jjl014: Yup!

PR #53, opened 2020-08-18
Update trackPurchase method to point to V2 endpoint.

sblaurock:
✅  219 passing


sblaurock:
> Need to talk with DS about `order_id`, `section`

@rubikzube `section` is restored - `order_id` doesn't seem to be supported by backend currently, but I'll dig and see if that's important to get added. We can follow up with that once it's supported.

PR #55, opened 2020-08-26
Add support for passing order_id to trackPurchase method.

sblaurock:
✅  All tests pass.


spec/src/modules/tracker.js
```diff
@@ -1282,7 +1282,10 @@ describe('ConstructorIO - Tracker', () => {
       ],
       revenue: 123.45,
     };
-    const optionalParameters = { section: 'Products' };
+    const optionalParameters = {
+      order_id: '123938123',
```
jjl014, line 1286: Should we check for this in the optional params test we have below?

sblaurock: > Should we check for this in the optional params test we have below?

Nice catch! Thanks

PR #59, opened 2020-09-10
[CH11185] Disable dispatch of tracking events by default

sblaurock:
✅  All tests pass (225/225)

- `sendTrackingEvents` is now defaulted to false within request queue, meaning we will not dispatch events from the library when default options are supplied

- Changed logic so that events can still be queued (stored in localstorage) but will never be sent if `sendTrackingEvents` is not true

- Moved options relating to request queue into an options object rather than having them as individual options

- Updated all tests

Easier to view diff with [?w=1](https://github.com/Constructor-io/constructorio-client-javascript/pull/59/files?w=1)

### DIFF
--- a/src/constructorio.js
+++ b/src/constructorio.js
@@ -25,6 +25,9 @@ class ConstructorIO {
    * @param {function} [fetch] - If supplied, will be utilized for requests rather than default Fetch API
    * @param {number} [trackingSendDelay] - Amount of time to wait before sending tracking events (in ms)
    * @param {boolean} [sendTrackingEvents] - Indicates if tracking events should be dispatched
+   * @param {object} [requestQueue] - Options related to 'RequestQueue' class
+   * @param {boolean} [requestQueue.sendTrackingEvents] - Indicates if tracking events should be dispatched
+   * @param {boolean} [requestQueue.trackingSendDelay] - Amount of time to wait before sending tracking events (in ms)
    * @param {object} [idOptions] - Options object to be supplied to 'constructorio-id' module
    * @param {object} [eventDispatcher] - Options related to 'EventDispatcher' class
    * @param {boolean} [eventDispatcher.enabled] - Determine if events should be dispatched
@@ -47,8 +50,7 @@ class ConstructorIO {
       sessionId,
       userId,
       fetch,
-      trackingSendDelay,
-      sendTrackingEvents,
+      requestQueue,
       eventDispatcher,
       idOptions,
     } = options;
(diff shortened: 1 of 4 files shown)


src/utils/request-queue.js
```diff
@@ -15,21 +15,25 @@ class RequestQueue {
...
-    this.send();
+    if (this.sendTrackingEvents) {
+      this.send();
+    }
   }
 
   // Add request to queue to be dispatched
   queue(url, method = 'GET', body) {
-    if (this.sendTrackingEvents && !this.humanity.isBot()) {
+    if (!this.humanity.isBot()) {
```
rubikzube, line 36: There may be legal implications here that are problematic... we should probably revert this change

src/utils/request-queue.js
```diff
@@ -43,87 +47,93 @@ class RequestQueue {
...
-            body: JSON.stringify(nextInQueue.body),
-            mode: 'cors',
-            headers: { 'Content-Type': 'text/plain' },
-          });
-        }
-
-        if (request) {
-          this.requestPending = true;
-          const instance = this;
+    if (this.sendTrackingEvents) {
```
rubikzube, line 50: There may be legal implications here that are problematic... we should probably revert this change

src/constructorio.js
```diff
@@ -47,8 +48,7 @@ class ConstructorIO {
       sessionId,
       userId,
       fetch,
-      trackingSendDelay,
-      sendTrackingEvents,
+      requestQueue,
```
rubikzube, line 51: If we are aiming for speed, let's keep the existing interface as changing the name will result in changes in other repos also.  If we are aiming for correctness, maybe rename to `requestQueue` to `trackingOptions`?

rubikzube (changes requested):

src/utils/request-queue.js
sblaurock: > There may be legal implications here that are problematic... we should probably revert this change

Resolved

src/constructorio.js
sblaurock: > If we are aiming for speed, let's keep the existing interface as changing the name will result in changes in other repos also. If we are aiming for correctness, maybe rename to `requestQueue` to `trackingOptions`?

I figure we need to make a change anyway to explicitly turn on `sendTrackingEvents`, so it would be a good time to do this refactor now.

>  If we are aiming for correctness, maybe rename to requestQueue to trackingOptions

The event emitter options are currently named after the class - was following the same pattern. trackingOptions works for me, though it's a bit odd that idOptions is separate from this.

jjl014 (approved):
nice, lookin' good! 🙂

rubikzube (approved):
Let's do it/

PR #64, opened 2020-09-18
[CH 11170] Beacon Mode

jjl014:
#### Updates:
* Send along `beacon: true` in body params of POST requests (mostly applicable to click events and purchase)
* Add tests

#### Context:
* Need this so that if we send a post request with an item id that's not found in the database, we won't treat it as a bad request and still log it in kibana so we have it for unknown items and debugging


src/modules/tracker.js
```diff
@@ -79,6 +85,7 @@ function applyParamsAsString(parameters, options) {
 class Tracker {
   constructor(options) {
     this.options = options;
+    this.options.beaconMode = (options && options.beaconMode === false) ? false : true; // Defaults to 'true'
```
rubikzube, line 88: I would be careful here modifying passed in parameters ... should probably set `this.options = {...options}`

rubikzube (changes requested):

PR #72, opened 2020-12-08
[CH 11952] Expose Purchase Event Helper Functions

jjl014:
#### Updates:
* Expose helper functions for checking/adding order ids in the storage
* Will be class level methods (i.e. usage `Tracker.addOrderIdRecord` and `Tracker.hasOrderIdRecord`)
* Doesn't get built into the docs 👍

### DIFF
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -418,11 +418,11 @@ class Tracker {
 
       if (order_id) {
         // Don't send another purchase event if we have already tracked the order
-        if (Tracker.hasOrderIdRecord(order_id)) {
+        if (this.hasOrderIdRecord(order_id)) {
           return false;
         }
 
-        Tracker.addOrderIdRecord(order_id);
+        this.addOrderIdRecord(order_id);
 
         // Add order_id to the tracking params
         bodyParams.order_id = order_id;
@@ -864,7 +864,7 @@ class Tracker {
   /**
    * Expose helper function to add order id to storage
    */
-  static addOrderIdRecord(orderId) {
+  addOrderIdRecord(orderId) { /* eslint-disable-line class-methods-use-this */
     let purchaseEventStorage = JSON.parse(store.session.get(purchaseEventStorageKey));
     const orderIdHash = CRC32.str(orderId.toString());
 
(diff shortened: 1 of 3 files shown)


src/modules/tracker.js
```diff
@@ -856,6 +860,47 @@ class Tracker {
...
+  addOrderIdRecord(orderId) { /* eslint-disable-line class-methods-use-this */
+    let purchaseEventStorage = JSON.parse(store.session.get(purchaseEventStorageKey));
+    const orderIdHash = CRC32.str(orderId.toString());
+
+    if (purchaseEventStorage) {
+
+      // If the order already exists, do nothing
+      if (purchaseEventStorage[orderIdHash]) {
+        return;
+      }
```
deomsj, line 876: It's not hurting anything, but these lines can be removed...

```js
      // If the order already exists, do nothing
      if (purchaseEventStorage[orderIdHash]) {
        return;
      }
```

src/modules/tracker.js
```diff
@@ -1,9 +1,13 @@
 /* eslint-disable object-curly-newline, no-underscore-dangle, camelcase, no-unneeded-ternary */
 const qs = require('qs');
 const EventEmitter = require('events');
+const CRC32 = require('crc-32');
```
deomsj, line 4: nice :)

src/modules/tracker.js
```diff
@@ -856,6 +860,47 @@ class Tracker {
...
+   */
+  hasOrderIdRecord(orderId) { /* eslint-disable-line class-methods-use-this */
+    const purchaseEventStorage = JSON.parse(store.session.get(purchaseEventStorageKey));
+    const orderIdHash = CRC32.str(orderId.toString());
+
+    if (purchaseEventStorage && purchaseEventStorage[orderIdHash]) {
+      return true;
+    }
+
+    return null;
```
deomsj, line 902: To me, `null` is used to explicitly communicates that there is no value to return, but I would expect `false` to be returned by a function named `hasOrderIdRecord` if the `orderIdRecord` does not exist. Either return value is falsey, but using `false` instead of `null` here gets our helper function to the happy place of always returning a boolean value.

deomsj (approved):
nice! left a few thoughts in comments, but code looks great to me

PR #73, opened 2020-12-16
Expose fmtOptions for search and browse

shivanetd:


src/modules/browse.js
```diff
@@ -81,6 +81,11 @@ function createBrowseUrl(filterName, filterValue, parameters, options) {
     if (section) {
       queryParams.section = section;
     }
+
+    // Pull ftm options from parameters
```
sblaurock, line 85: Small typo here - could we change to `Pull format options from parameters`?

src/modules/browse.js
```diff
@@ -116,6 +121,7 @@ class Browse {
    * @param {object} [parameters.filters] - Filters used to refine results
    * @param {string} [parameters.sortBy='relevance'] - The sort method for results
    * @param {string} [parameters.sortOrder='descending'] - The sort order for results
+   * @param {object} [parameters.fmtOptions] - fmtOptions used to refine groups
```
sblaurock, line 124: What do you think about `The format options used to refine result groups`?

src/modules/search.js
```diff
@@ -82,6 +82,11 @@ function createSearchUrl(query, parameters, options) {
     if (collectionId) {
       queryParams.collection_id = collectionId;
     }
+
+    // Pull ftm options from parameters
```
sblaurock, line 86: Small typo here - could we change to `Pull format options from parameters`?

src/modules/search.js
```diff
@@ -116,6 +121,7 @@ class Search {
    * @param {object} [parameters.filters] - Filters used to refine search
    * @param {string} [parameters.sortBy='relevance'] - The sort method for results
    * @param {string} [parameters.sortOrder='descending'] - The sort order for results
+   * @param {object} [parameters.fmtOptions] - fmtOptions used to refine groups
```
sblaurock, line 124: What do you think about The format options used to refine result groups?

sblaurock (commented):
✅  Lint passes
✅  Tests pass (2 added)

This looks great to me - thank you @shivanetd! Love the tests - much appreciated.

A few minor comments attached. Once we've taken care of those, I'd be happy to merge and release this.

spec/src/modules/browse.js
```diff
@@ -184,6 +184,27 @@ describe('ConstructorIO - Browse', () => {
...
+
+      browse.getBrowseResults(filterName, filterValue, { fmtOptions }).then((res) => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+
+        expect(res).to.have.property('request').to.be.an('object');
+        expect(res).to.have.property('response').to.be.an('object');
+        expect(res).to.have.property('result_id').to.be.an('string');
+        expect(res.request.fmt_options).to.deep.equal(fmtOptions);
+        expect(requestedUrlParams).to.have.property('fmt_options');
+        expect(requestedUrlParams.fmt_options).to.have.property('groups_max_depth').to.equal(Object.values(fmtOptions)[0][0]);
```
sblaurock, line 202: This test does not pass, unfortunately. Try:

```
expect(requestedUrlParams.fmt_options).to.have.property('groups_max_depth').to.equal(Object.values(fmtOptions)[0].toString());
```

The reason being that we do not need the second `[0]` accessor, and the query parameter passed is of type string, not integer as defined in the `fmtOptions` object.

spec/src/modules/search.js
```diff
@@ -193,6 +193,29 @@ describe('ConstructorIO - Search', () => {
...
+        fmtOptions,
+      }).then((res) => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+
+        expect(res).to.have.property('request').to.be.an('object');
+        expect(res).to.have.property('response').to.be.an('object');
+        expect(res).to.have.property('result_id').to.be.an('string');
+        expect(res.request.fmt_options).to.deep.equal(fmtOptions);
+        expect(requestedUrlParams).to.have.property('fmt_options');
+        expect(requestedUrlParams.fmt_options).to.have.property('groups_max_depth').to.equal(Object.values(fmtOptions)[0][0]);
```
sblaurock, line 214: This test does not pass, unfortunately. Try:

```
expect(requestedUrlParams.fmt_options).to.have.property('groups_max_depth').to.equal(Object.values(fmtOptions)[0].toString());
```

The reason being that we do not need the second `[0]` accessor, and the query parameter passed is of type string, not integer as defined in the `fmtOptions` object.

sblaurock (changes requested):

PR #76, opened 2020-12-28
Ensure library can be imported and utilized safely within a Node.js context

sblaurock:
- Add tests to ensure invocation succeeds without a `window` object
- Move methods that rely on `window` into helpers and safeguard
- Disable tracking / event dispatching if no window present
- Don't inherit session and client ID's from identity module if no window present
- All tests pass (273/273)


crgee1 (approved):
tests pass, lgtm

rubikzube (approved):

PR #78, opened 2021-02-22
[CH 12884] Add support for variation id in search click through event

jjl014:
#### Updates:
* Add variation id support in search click through
* Add test


deomsj (commented):
I will leave for another review from folks who know more about this repo, but the code makes sense and looks good to me!

crgee1 (approved):
lgtm

PR #80, opened 2021-03-22
[CSL 8] Add Conversion Types to Tracking

jjl014:
#### Updates:
* Update trackConversion to use v2 endpoint
* Add support for converstion types
* Add/Update tests

### DIFF
--- a/spec/src/modules/search.js
+++ b/spec/src/modules/search.js
@@ -37,7 +37,6 @@ describe('ConstructorIO - Search', () => {
   describe('getSearchResults', () => {
     const query = 'drill';
     const section = 'Products';
-    const collectionId = 'test';
 
     it('Should return a response with a valid query, and section', (done) => {
       const { search } = new ConstructorIO({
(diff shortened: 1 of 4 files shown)


src/modules/tracker.js
```diff
@@ -356,37 +356,66 @@ class Tracker {
...
+      if (is_custom_type) {
+        bodyParams.is_custom_type = is_custom_type;
+      }
+
+      if (display_name) {
+        bodyParams.display_name = display_name;
       }
 
-      this.requests.queue(`${url}${applyParamsAsString(queryParams, this.options)}`);
+      const requestURL = `${requestPath}${applyParamsAsString({}, this.options)}`;
```
crgee1, line 410: are we not adding the item_name and item_id still to the request url?

crgee1 (changes requested):
looks cool, left one comment, and if you could update the docs for the new params

src/modules/tracker.js
jjl014: Updated to include it 👍

spec/src/modules/search.js
```diff
@@ -292,24 +291,6 @@ describe('ConstructorIO - Search', () => {
       });
     });
 
-    it('Should return a response from collection with a valid query and collection id', (done) => {
```
deomsj, line 295: why are we removing these collection tests? Are these no longer supported?

deomsj:
Is it an accurate summary to say that when tracking conversion events that have `is_custom_type: true`:
 - `display_name` is required
 - `type` is optional

crgee1 (approved):
lgtm

spec/src/modules/search.js
jjl014: Yea, the search endpoint no longer supports collection ids. It just returns an error:

`"Search within a collection is not supported."`

jjl014:
> Is it an accurate summary to say that when tracking conversion events that have is_custom_type: true:
> * display_name is required
> * type is optional 

@deomsj Yup, that is correct. Although I'm not entirely sure what happens in that case. (i.e. what is used for the for the type? does it default to `add_to_cart`?). Will followup w/ Wilson to get more details.

sblaurock:
@jjl014 will this break any existing implementations that we have in beacons for customers?

PR #86, opened 2021-04-12
Revert lockfile to version 1, update clean command.

sblaurock:
#85 will force us to be using the current Node LTS before versioning this package, which uses an npm version with lockfile version 1. This is preferable for now until we upgrade to the next LTS.

All tests pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Constructor.io JavaScript client",
   "main": "lib/constructorio.js",
   "scripts": {
-    "clean": "sudo rm -rf node_modules package-lock.json && npm install",
+    "clean": "sudo rm -rf node_modules package-lock.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Constructor.io JavaScript client",
   "main": "lib/constructorio.js",
   "scripts": {
-    "clean": "rm -rf node_modules",
+    "clean": "sudo rm -rf node_modules package-lock.json && npm install",
```
jjl014, line 7: Should we leave out `npm install` here to be consistent with the other repos?

jjl014 (commented):
👍

sblaurock:
@jjl014 updated - great catch

jjl014 (approved):
🚢

PR #93, opened 2021-05-11
Better encoding handling

jjl014:
#### Updates:
* Fixes an issue where there are brackets or other special characters used in the query params
* This was specifically affecting Sephora
    * https://www.sephora.com/shop/moisturizer-skincare?ref=filters[Brand]=Algenist,filters[Ingredient%20Preferences]=Vegan
    * Going to the URL above, you will see that when sending the browse results load event, it errors out on our `ourEncodeURIComponent` because it's origin_referrers that have brackets in it incorrectly.
    * Calling `encodeURIComponent` before passing it to qs.parse resolves this issue and should also encode any other special characters
    
Screen Shot 2021-05-11 at 1 20 33 PM

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -8,7 +8,7 @@ const purchaseEventStorageKey = '_constructorio_purchase_order_ids';
 const utils = {
   ourEncodeURIComponent: (str) => {
     if (str) {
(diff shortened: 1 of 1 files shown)


esezen (approved):
🚀

PR #96, opened 2021-05-21
[CSL-339] Add browse items endpoint and tests

esezen:


esezen:
Convert ids to itemIds

spec/src/modules/browse.js
```diff
@@ -634,18 +634,24 @@ describe('ConstructorIO - Browse', () => {
       browse.getBrowseResultsByItemIds(ids);
     });
 
-    it('Should be rejected when invalid ids is provided', () => {
+    it('Should be rejected when invalid ids are provided', () => {
```
esezen, line 637: ids technically 'is' an array. That's why I kept it as is but this reads better.

sblaurock (approved):
Fantastic - great work here. Made a few tweaks and will get it shipped.

PR #98, opened 2021-06-04
Psl 111 tracking generic result click

deomsj:
Add generic result click to the client trackers (already exists in the back end behavioral API v2)
 - This event should be sent exclusively from other click events
 - This event should be fired whenever a user clicks on a detail page HREF
 - Add event to constructorio-client-javascriptwire in through autocomplete-ui
 - Event properties
    - item_id: required
    - item_name: optional
    - variation_id: optional
    - section: optional, defaults to 'products'

Screen Shot 2021-06-04 at 8 25 03 PM

Note: tests are not passing consistently for me, but this is also true for other methods... is this a known issue? Any idea what could be causing this?


spec/src/modules/tracker.js
```diff
@@ -3667,6 +3667,344 @@ describe('ConstructorIO - Tracker', () => {
     });
   });
 
+
+  describe('trackGenericResultClick', () => {
+    // Note: `variation_id` parameter not being passed as none are defined
```
jjl014, line 3672: What does `parameter not being passed as none are defined` mean? 🤔

spec/src/modules/tracker.js
```diff
@@ -3667,6 +3667,344 @@ describe('ConstructorIO - Tracker', () => {
...
+        // Response
+        expect(eventSpy).to.have.been.called;
+        expect(responseParams).to.have.property('method').to.equal('POST');
+        expect(responseParams).to.have.property('message');
+
+        done();
+      }, waitInterval);
+    });
+
+    it('Should respond with a valid response when required and optional parameters are provided', (done) => {
```
jjl014, line 3864: I think this test description needs an update.

spec/src/modules/tracker.js
```diff
@@ -3667,6 +3667,344 @@ describe('ConstructorIO - Tracker', () => {
...
+        // Response
+        expect(eventSpy).to.have.been.called;
+        expect(responseParams).to.have.property('method').to.equal('POST');
+        expect(responseParams).to.have.property('message').to.equal('ok');
+
+        done();
+      }, waitInterval);
+    });
+
+    it('Should respond with a valid response when required and optional parameters are provided', (done) => {
```
jjl014, line 3899: Same as above, test description needs an update?

jjl014 (commented):
Nice work, this looks great! Just left a couple comments on the tests. 👍

sblaurock:
Don't check in changes to `./docs` please! These will get compiled with the `version` command.

spec/src/modules/tracker.js
sblaurock: `variation_id` is a valid parameter that can be passed to this method, however none are currently defined in the catalog we use for testing, so we aren't ever passing it in testing

This comment definitely could be more clear - I'll update in this PR. Good callout @jjl014 !

spec/src/modules/tracker.js
sblaurock: I think this isn't the only place that this mistake occurs - it's my fault originally!

sblaurock (approved):
Looks great to me! Nice work @deomsj. I made a few changes to help with the comments Jimmy raised - should be good to go. I'll merge and get this shipped 🚀

spec/src/modules/tracker.js
sblaurock: I've updated these

deomsj:
awesome, thanks for the reviews and help getting this one across the finish line @jjl014  and @sblaurock!!!

PR #103, opened 2021-06-24
Remove waiting to check results of tracking tests.

sblaurock:
Tracking tests are notoriously long to run and experience a lot of false negatives, as we're waiting for a set duration in order to check the status of the test.

This PR removes the wait, and simply checks the assertions once the "success" event es emitted. The result is:

- Less code
- Much, much faster test runs (about five times as fast)
- Fewer incidences of false positives

Before (8 failures):
```
npm run test  4.98s user 0.61s system 4% cpu 2:18.70 total
```

After (0 failures):
```
npm run test  4.28s user 0.61s system 17% cpu 28.375 total
```

### DIFF
--- a/spec/src/utils/request-queue.js
+++ b/spec/src/utils/request-queue.js
@@ -14,9 +14,10 @@ dotenv.config();
 
 describe('ConstructorIO - Utils - Request Queue', () => {
   const storageKey = '_constructorio_requests';
-  const waitInterval = 700;
+  const waitInterval = 1000;
   const requestQueueOptions = {
     sendTrackingEvents: true,
+    trackingSendDelay: 1,
   };
 
   describe('queue', () => {
--- a/spec/src/utils/request-queue.js
+++ b/spec/src/utils/request-queue.js
@@ -23,6 +23,10 @@ describe('ConstructorIO - Utils - Request Queue', () => {
   describe('queue', () => {
     let defaultAgent;
 
+    before(() => {
+      helpers.clearStorage();
+    });
+
     beforeEach(() => {
       global.CLIENT_VERSION = 'cio-mocha';
       helpers.setupDOM();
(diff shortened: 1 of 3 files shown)


esezen (commented):
This looks great! The tests run way faster. There are 9 tests (all from utils) failing for me. Is this normal? Are we going to handle those separately?

sblaurock:
> This looks great! The tests run way faster. There are 9 tests (all from utils) failing for me. Is this normal? Are we going to handle those separately?

@esezen good catch! I'll investigate those as well.

sblaurock:
@esezen should be all set now!

spec/src/utils/request-queue.js
```diff
@@ -14,9 +14,10 @@ dotenv.config();
 
 describe('ConstructorIO - Utils - Request Queue', () => {
   const storageKey = '_constructorio_requests';
-  const waitInterval = 700;
+  const waitInterval = 1000;
```
esezen, line 17: Awesome! For my own learning, can you explain how this fixes the issue with utils tests?

esezen (approved):
Great work! 🚀

spec/src/utils/request-queue.js
sblaurock: It's actually the other line change that fixes the problem: 

```
trackingSendDelay: 1,
```

We recently changed the trackingSendDelay default from `150` (ms) to `250`. Since the test wait interval was 700, and some of the test required dispatching three events, it would sometimes take longer than the 700ms wait interval (250 * 3 = 750).

Changing the send delay to something very short ensures that we have plenty of time for the request queue to "clear" before checking the results. The change from `700` to `1000` was basically made just because I like "round" numbers 😛

esezen: Thank you!

PR #109, opened 2021-07-23
[CSL-215] Java SDK : Remove API key and token

esezen:
The changes in the Java tests and the catalog CSV files we use broke two tests in this repo. This PR fixes them and uses the same story for reference


crgee1 (approved):
This doesn't look like it has much to do with keys or tokens but I also don't see a problem with these changes

esezen:
> This doesn't look like it has much to do with keys or tokens but I also don't see a problem with these changes

I should have added an explanation. Sorry about that. Added it now

PR #110, opened 2021-07-23
[CSL-473] Update all client-javascript / client-node to use latest of identity module

esezen:
All tests except 2 pass. Those are fixed in #109


crgee1 (approved):
lgtm

PR #111, opened 2021-07-26
[CSL-477] Add support for hiddenFields

esezen:


spec/src/modules/browse.js
```diff
@@ -607,6 +626,24 @@ describe('ConstructorIO - Browse', () => {
...
+        fetch: fetchSpy,
+      });
+
+      browse.getBrowseResultsByItemIds(ids, { hiddenFields }).then((res) => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+
+        expect(res).to.have.property('request').to.be.an('object');
+        expect(res).to.have.property('response').to.be.an('object');
+        expect(res).to.have.property('result_id').to.be.an('string');
+        expect(requestedUrlParams).to.have.property('hidden_fields').to.eql(hiddenFields);
```
jjl014, line 642: Should we have a

```expect(res.request.hidden_fields).to.eql(hiddenFields);```

check here as well?

jjl014 (commented):
Nice, looking good to me!

Just left one comment around the tests.

spec/src/modules/browse.js
esezen: Yes. Thanks for catching it!

jjl014 (approved):

PR #116, opened 2021-07-31
[CSL-491] Concept consistency improvements

sblaurock:
- Any method that used to accept `customer_id` will still do so, but is documented as `item_id` to be more consistent + tests
- Any method that used to accept `name` will still do so, but is documented as `item_name` to be more clear and consistent + tests
- More concise documentation of concepts globally across tracking module

This change does not introduce any breakage to existing implementations / previous versions. We'll need to port these changes to `constructorio-node` once approved.


esezen (approved):
LGTM! 🚀

PR #122, opened 2021-08-27
[CSL-543] Migrate timeout functionality from node to client-javascript

esezen:
These tests sometimes fail:
image

### DIFF
--- a/spec/mocha.helpers.js
+++ b/spec/mocha.helpers.js
@@ -9,6 +9,7 @@ const setupDOM = () => {
 
   global.window = window;
   global.document = window.document;
+  global.AbortController = window.AbortController;
 };
 
 // Tear down mock DOM environment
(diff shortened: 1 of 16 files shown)


jjl014 (approved):
Nice! This looks great to me!

There are a couple tests towards the end that seems to fail randomly, but I was able to get a couple successful runs so probably not a big deal?

sblaurock:
> Nice! This looks great to me!
> 
> There are a couple tests towards the end that seems to fail randomly, but I was able to get a couple successful runs so probably not a big deal?

This happens to me on occasion as well - I think somehow the localRtorage values can persist across test runs. I'll take it as an item to look into, but I'm pretty confident it's not relevant to this PR.

sblaurock (approved):
Great work on this!

PR #123, opened 2021-09-02
[CSL-517] Research best SDK approach for React Native (and other DOM-less environments)

sblaurock:
- If DOM context is not available, require client and session identifier to be set on instantiation

- Update DOM context detection method - inspiration from: https://stackoverflow.com/questions/32216383/in-react-how-do-i-detect-if-my-component-is-rendering-from-the-client-or-the-se

- Set sendTrackingEvents to `true` by default if no DOM context is available

- Update tests

### DIFF
--- a/spec/src/constructorio.js
+++ b/spec/src/constructorio.js
@@ -94,6 +94,15 @@ describe('ConstructorIO', () => {
     expect(() => new ConstructorIO()).to.throw('API key is a required parameter of type string');
   });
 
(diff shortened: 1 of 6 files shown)


esezen (commented):
I tested with React Native and everything seems to be working. I also ran autocomplete-ui tests using this branch and they are passing

rubikzube (approved):

PR #129, opened 2021-09-21
[CSL-626] Bundled distributable

sblaurock:
Bundles entire library as single JS file to be included within frontends without a `npm` powered build process.

Exposed in `./dist`

- [x] Run entire test suite against bundled distributable
- [x] Add client version indicator to bundled distributable (`-bundled-`) + tests
- [x] Update readme
- [x] Add header to output file (use `search-ui` for inspiration)


esezen (approved):
LGTM!

PR #135, opened 2021-11-01
Add link to DOM-less environment information on wiki.

sblaurock:


jjl014 (approved):
Send that informatation to the moon! 🌕  🚀

PR #147, opened 2021-11-19
[CSL 731] Util Tests

jjl014:
### Updates:
* Add tests around util helpers
* Bring in `@babel/plugin-transform-runtime` to be able to use `async`/`await` in tests
* All tests pass
* Not 100% coverage, but I don't think it's possible to test the catch block (screenshot below) of `createCustomEvent` without actually passing a variable that hasn't been defined yet which would fail the catch statement too.
    * Looks like CustomEvent can take a bunch of different things, even empty objects, null, etc.
* Update `request-queue` spec to point to `test` directory

Screen Shot 2021-11-22 at 4 19 19 PM
Screen Shot 2021-11-22 at 4 21 00 PM

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,4 @@
 {
-  "presets": ["@babel/preset-env"]
+  "presets": ["@babel/preset-env"],
+  "plugins": ["@babel/plugin-transform-runtime"]
 }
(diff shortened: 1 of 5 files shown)


esezen (approved):
LGTM! 🚀 Great job

esezen (approved):
🚀🚀🚀

sblaurock:
`should` to `Should`

PR #150, opened 2021-12-27
[CSL-806] Investigate duplicate events - Use sendBeacon

esezen:
How to test:

1. Change source of clientJS in your autocomplete-ui `package.json` to local path: `    "@constructor-io/constructorio-client-javascript": "/Users/XXX/constructorio-client-javascript"`
2. In clientJS repo, do `npm run compile`
3. In autocomplete-ui repo, do `npm run build:dev:all`


sblaurock:
Closing for now - we will revisit sendBeacon approach once we're on v2 API (all requests are POSTS) so that we can implement globally.

PR #151, opened 2021-12-27
[CSL-806] Investigate duplicate events - Use indexeddb

esezen:
Alternative solution. Uses indexeddb instead of localStorage. Adds less than 1kb to bundle size


src/utils/request-queue.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable brace-style, no-unneeded-ternary */
 const fetchPonyfill = require('fetch-ponyfill');
 const Promise = require('es6-promise');
-const store = require('../utils/store');
+const idb = require('idb-keyval/dist/compat');
```
esezen, line 4: IE10 compatible

spec/mocha.helpers.js
```diff
@@ -10,6 +12,9 @@ const setupDOM = () => {
   global.window = window;
   global.document = window.document;
   global.AbortController = window.AbortController;
+  global.navigator = { indexedDB };
+  global.indexedDB = indexedDB;
+  idb.clear();
```
esezen, line 17: This needs to change

src/utils/request-queue.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable brace-style, no-unneeded-ternary */
 const fetchPonyfill = require('fetch-ponyfill');
 const Promise = require('es6-promise');
-const idb = require('idb-keyval/dist/compat');
+const idb = require('idb-keyval');
```
esezen, line 4: We should figure out how to make tests work with compat version

sblaurock:
Closing as switching to synchronous data storage pattern has large implementation implications.

PR #152, opened 2021-12-28
Add preventative measure to ensure queue is set adequately.

sblaurock:
- If queue length is not set correctly, remove queue entirely.
- Ensure queue length cannot exceed 20 items.


esezen (approved):
LGTM!

PR #155, opened 2021-12-29
Do not emit error event on queue flush for expired request items.

sblaurock:


esezen (approved):
Tests pass and beacon works

PR #163, opened 2022-01-20
[CSL-884] Investigate alternative to pre-push

esezen:
To test:
* Create a lint issue in code
* Try pushing to this branch
* Push should fail


crgee1 (approved):
lgtm

.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
jjl014, line 2: Just for my curiousity. What does this do?

jjl014 (approved):
Nice, lgtm!

.husky/pre-push
esezen: Not sure tbh. It is automatically created by husky per hook (push, commit etc).

https://github.com/typicode/husky
```
npx husky add .husky/pre-commit "npm test"
git add .husky/pre-commit
```

jjl014: Oooo, interesting. And I'm guessing you just add any npm commands you want to be ran pre-push into this file?

esezen: Yessir

sblaurock:
Works great!

sblaurock:
Can we port to constructorio-node as well?

esezen:
> Can we port to constructorio-node as well?

Hmm I thought I did but it looks like I missed that one. Will do

PR #164, opened 2022-01-28
[CSL-942] Update dependencies to avoid security warnings

esezen:
Fixes everything except `shelljs` which is fixed in #163 
Tests pass.


jjl014 (approved):
🚢 🚀

PR #173, opened 2022-04-04
[CSL-1124] Resolve process is undefined issue

esezen:


sblaurock (approved):

src/constructorio.js
```diff
@@ -20,7 +20,7 @@ const computePackageVersion = () => {
     versionModifiers.push('domless');
   }
 
-  if (typeof process.env !== 'undefined' && process.env.BUNDLED) {
+  if (typeof process !== 'undefined' && typeof process.env !== 'undefined' && process.env.BUNDLED) {
```
KeysiJones, line 23: You can do this to avoid repetitive "typeof" comparisons

if (process?.env?.BUNDLED) {

}

It has the same behavior as your code, but it's cleaner.

PR #177, opened 2022-04-13
Encoding refactor.

sblaurock:
We should wait to merge this to make sure we don't cause issues - this is a breaking change potentially.

- Simplify URL encoding of query parameters
- Do not encode body parameters
- Add tests to ensure stability around the above


spec/src/modules/tracker.js
```diff
@@ -311,6 +311,61 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
+        expect(responseParams).to.have.property('message').to.equal('ok');
+
+        done();
+      });
+
+      expect(tracker.trackSessionStart()).to.equal(true);
+    });
+
+    it('Should properly transform non-breaking spaces in parameters', (done) => {
+      const breakingSpaces = '   ';
```
esezen, line 343: Should this be nonBreakingSpaces?

esezen:
One test fails. We can discuss how we want to handle that
Screen Shot 2022-04-15 at 4 25 58 PM
.

esezen (approved):
LGTM!

PR #184, opened 2022-06-28
[CSL 1397] Change item_detail_load to v2

crgee1:


esezen (changes requested):
Changes look great but some of the `trackItemDetailLoad` tests fail for me 🤔 
Also do we know if DS needs to something on their end since we are moving from V1 to V2?

crgee1:
> Changes look great but some of the `trackItemDetailLoad` tests fail for me 🤔 Also do we know if DS needs to something on their end since we are moving from V1 to V2?

No, they just need it to be v2

esezen (approved):
LGTM!

PR #185, opened 2022-06-29
[CSL-1396] Add variations mapping API support

esezen:


stanlp1 (approved):
We have merge conflicts. And 2 tests for hiddenFields fail but I think thats unrelated.

Didn't know that recommendations also support variation_maps o.o

PR #187, opened 2022-06-30
fixed 2 failing hidden fields/facet tests

Here2Huynh:


spec/src/modules/search.js
```diff
@@ -298,14 +298,17 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
       });
 
       search.getSearchResults(query, { section, hiddenFields }, {}).then((res) => {
+
```
stanlp1, line 301: Extra Newline

spec/src/modules/search.js
```diff
@@ -298,14 +298,17 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
...
 
+        const resultWithTestField = res.response.results.find(result => result.data.testField);
+
         expect(res).to.have.property('request').to.be.an('object');
         expect(res).to.have.property('response').to.be.an('object');
         expect(res).to.have.property('result_id').to.be.an('string');
         expect(res.request.fmt_options.hidden_fields).to.eql(hiddenFields);
         expect(requestedUrlParams.fmt_options).to.have.property('hidden_fields').to.eql(hiddenFields);
-        expect(res.response.results[0].data).to.have.property('testField').to.eql('hiddenFieldValue');
+        expect(resultWithTestField).to.be.not.null.to.eql('hiddenFieldValue');
```
stanlp1, line 311: I think we actually need to change this. res.response.results.find returns undefined when nothing is found. And undefined !== null.

We will probably have to use something else other than to.be.not.null because that will always be true at the moment.

spec/src/modules/search.js
```diff
@@ -317,15 +320,17 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
...
 
+        const facetWithNameBrand = res.response.facets.find(facet => facet.name === 'Brand');
+
         expect(res).to.have.property('request').to.be.an('object');
         expect(res).to.have.property('response').to.be.an('object');
         expect(res).to.have.property('result_id').to.be.an('string');
         expect(res.request.fmt_options.hidden_facets).to.eql(hiddenFacets);
         expect(requestedUrlParams.fmt_options).to.have.property('hidden_facets').to.eql(hiddenFacets);
-        expect(res.response.facets[0]).to.have.property('name').to.eql('Brand');
+        expect(facetWithNameBrand).to.be.not.null;
```
stanlp1, line 333: Same for here, this will always be true at the moment.

stanlp1 (changes requested):
Nice work 👍 , we got the right idea but there are some small issues, due to the intricacies of java script 🙃

spec/src/modules/search.js
Here2Huynh: Got it, added more precise tests instead.

spec/src/modules/search.js
Here2Huynh: Good eye, added better tests for that case.

PR #188, opened 2022-06-30
Csl 1411 client javascript investigate replacing

Here2Huynh:
https://linear.app/constructor/issue/CSL-1411/client-javascript-investigate-replacing-mocha-jsdom-with-jsdom-global


package.json
```diff
@@ -49,7 +49,8 @@
     "eslint-plugin-import": "^2.25.2",
     "husky": "^7.0.4",
     "jsdoc": "^3.6.7",
-    "jsdom": "^15.1.1",
+    "jsdom": "15.2.1",
+    "jsdom-global": "3.0.2",
```
esezen, line 53: It doesn't let me comment on line 58, can we remove `mocha-jsdom`?

PR #189, opened 2022-07-06
Incomplete regular expression for hostnames

mocca102:


stanlp1 (changes requested):
cool, codeql works for prs too!. Looks like we will need to make some extra changes for the periods.

Similar to the first issue with google bot, we are only escaping the periods once so we would want to escape those twice or else the periods would be treated as wildcards in regex. And lets also escape all the other periods too.

stanlp1 (approved):
LGTM. But lets check with Jimmy to make sure these periods were intended to be periods and not wildcards.

jjl014 (approved):
This looks good to me! 👍

PR #194, opened 2022-07-29
[CSL-1274] Add item_name support recommendation click

stanlp1:
Some users don't expose item_id so item_name can be used.


jjl014 (approved):
lgtm!

PR #201, opened 2022-08-29
[CSL-1610] Store order_ids in local storage instead.

stanlp1:
This changes the order_id to be stored in local storage instead of session storage so deduplication works across tabs.

Changed the order_id storage to be an array instead of object so that we can limit # of order_ids to 10 in order of time. 

Also fixes some failing tests.

### DIFF
--- a/spec/src/modules/autocomplete.js
+++ b/spec/src/modules/autocomplete.js
@@ -282,11 +282,13 @@ describe(`ConstructorIO - Autocomplete${bundledDescriptionSuffix}`, () => {
 
       autocomplete.getAutocompleteResults(query, { hiddenFields }, {}).then((res) => {
         const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+        const item = res.sections.Products.find((product) => product.data.testField);
+        const itemData = item && item.data;
 
         expect(res).to.have.property('request').to.be.an('object');
         expect(res).to.have.property('sections').to.be.an('object');
         expect(res).to.have.property('result_id').to.be.an('string');
-        expect(res.sections.Products[0].data).to.have.property('testField').to.eql('hiddenFieldValue');
+        expect(itemData).to.have.property('testField').to.eql('hiddenFieldValue');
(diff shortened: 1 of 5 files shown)


spec/src/modules/autocomplete.js
```diff
@@ -282,11 +282,13 @@ describe(`ConstructorIO - Autocomplete${bundledDescriptionSuffix}`, () => {
...
       autocomplete.getAutocompleteResults(query, { hiddenFields }, {}).then((res) => {
         const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+        const item = res.sections.Products.find((product) => product.data.testField);
+        const itemData = item && item.data;
 
         expect(res).to.have.property('request').to.be.an('object');
         expect(res).to.have.property('sections').to.be.an('object');
         expect(res).to.have.property('result_id').to.be.an('string');
-        expect(res.sections.Products[0].data).to.have.property('testField').to.eql('hiddenFieldValue');
+        expect(itemData).to.have.property('testField').to.eql('hiddenFieldValue');
```
jjl014, line 291: Nice!

src/utils/helpers.js
```diff
@@ -113,28 +113,29 @@ const utils = {
...
     if (purchaseEventStorage) {
       // If the order already exists, do nothing
-      if (purchaseEventStorage[orderIdHash]) {
+      if (purchaseEventStorage.includes(orderIdHash)) {
         return;
       }
 
-      purchaseEventStorage[orderIdHash] = true;
+      if (purchaseEventStorage.length >= 10) {
+        purchaseEventStorage.slice(-9);
```
jjl014, line 129: I think this should be `purchaseEventStorage = purchaseEventStorage.slice(-9)` 🤔 

Probably not necessary, but should we add a test around this? 😛

jjl014 (commented):
Nice, looking good to me!

Left one comment 👍

jjl014 (approved):
Naisuu, lgtm!

PR #202, opened 2022-08-29
[NOCSL] Don't double stringify local storage data

stanlp1:
store2 already stringifies and parses data so I was double stringifying.


jjl014 (approved):
Nice catch! lgtm! 🚀

PR #204, opened 2022-10-04
Add babel runtime as peer dependency

jjl014:
### Updates:
* Add `@babel/runtime` as a peer dependency to resolve an issue with Vite (might also fix issues where the package doesn't exist already in customer's codebase)


pedro-lb (approved):
LGTM. Thanks for the work here! 💯

PR #206, opened 2022-10-07
[CSL 1710] Fix Bundle Tests

jjl014:
### Updates:
* `npm run test:bundled`
    * Fixed bundle tests by updating jsdom configuration [ All tests pass ]
    * Getting CORS errors which doesn't affect the tests 🤔. Not sure how to resolve it and wasn't able to find a helpful resource, could use some help here!
        * Seems to only happen on the tests where we use incorrect filter params in the result request tests which is interesting..

Screen Shot 2022-10-07 at 12 01 16 PM

### DIFF
--- a/spec/src/constructorio.js
+++ b/spec/src/constructorio.js
@@ -1,7 +1,6 @@
 /* eslint-disable no-unused-expressions, import/no-unresolved, no-new */
 const { expect } = require('chai');
 const sinon = require('sinon');
-const fs = require('fs');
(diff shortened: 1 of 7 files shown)


stanlp1 (approved):
Cors error only happens on the invalid filters because those request return 400s and don't have a access-control-allow-origin response header. No idea how to fix though 🤔 Weird it only happens in the bundled tests.

Tests pass looks good to me.

esezen (approved):
This is great! 🚀

PR #207, opened 2022-10-07
[CSL-1711] Beacon: Add `is_constructor_ab_cell` to all behavioral events

jesse-constructor:


spec/src/modules/tracker.js
```diff
@@ -68,20 +68,21 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
   });
 
   describe('trackSessionStart', () => {
-    it('Should respond with a valid response', (done) => {
+    it.only('Should respond with a valid response', (done) => {
```
jjl014, line 71: Let's remove the `.only` here

spec/src/modules/tracker.js
```diff
@@ -68,20 +68,21 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
       });
 
       tracker.on('success', (responseParams) => {
         const requestParams = helpers.extractUrlParamsFromFetch(fetchSpy);
-
         // Request
         expect(fetchSpy).to.have.been.called;
         expect(requestParams).to.have.property('key');
         expect(requestParams).to.have.property('i');
+        expect(requestParams).to.have.property('is_constructor_ab_cell');
```
jjl014, line 85: Can we also check the value of the property here?

jjl014 (commented):
Looks mostly good to me! Just left a couple of comments.

PR #210, opened 2022-10-27
Add support for Typescript types

mocca102:


sblaurock:
Will types be published with the bundle to `npm` as a result of this change? We likely don't want to ship this as part of that package as it will bloat the size.

stanlp1:
> Will types be published with the bundle to `npm` as a result of this change? We likely don't want to ship this as part of that package as it will bloat the size.

I'm not a typescript user but I _think_ that what we can do is publish to npm under the @types organization separately by making a pr in [this](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.

That way it remains optional if people want to include types, whether they want it as a dependency or dev dependency.

src/types/tests/search.test-d.ts
```diff
@@ -0,0 +1,386 @@
...
+			auto_generated_refined_query_rules: "default_rules",
+			manual_searchandizing: null,
+			personalization: "default_personalization",
+			filter_items: "filter_items_w_atcs_and_purchases",
+		},
+		searchandized_items: {},
+	},
+});
+
+// Example response from https://bonobos.com/search?term=red
```
stanlp1, line 143: I think we should probably remove references to customers here. So changing up the image links and what not.  Same for above.

sblaurock:
> `./dist` shouldn't be checked in I believe

@mocca102 just a reminder here - don't want to check in the dist file.

esezen:
This is looking good. We need to research how we want to include types. It would be nice to ship them with the package if they are not going to be included in the bundles and increase package size. If they will, we can consider Stanley's approach

package.json
```diff
@@ -20,7 +22,7 @@
     "coverage": "nyc --all --reporter=html npm run test:src",
     "postcoverage": "open coverage/index.html && rm -rf test",
     "docs": "jsdoc --configure ./.jsdoc.json ./README.md --recurse ./src --destination ./docs",
-    "compile": "babel src/ -d lib/",
+    "compile": "babel src/ -d lib/ --copy-files && rm -rf lib/types/tests",
```
sblaurock, line 25: @mocca102 can you explain the intent of `--copy-files` here?

sblaurock (commented):
This is looking great to me - one open question about `--copy-files`

sblaurock (approved):

PR #212, opened 2022-11-08
[CSL-1799] Investigate if we can remove es6-promise polyfill

esezen:


stanlp1 (approved):
Very promising.

jjl014 (approved):
Looking good to me. I only tested against Chrome & Firefox, but seems to be working just fine. Tested both Autocomplete UI and Search UI.

For others that want to test you can replace the current JS client in `package.json` with the following:

`"@constructor-io/constructorio-client-javascript": "file:../constructorio-client-javascript"` (might need to adjust the path a bit depending on your folder structures)

sblaurock (approved):

PR #213, opened 2022-11-10
[CSL-1813] Remove qs dependency

esezen:


src/utils/helpers.js
```diff
@@ -149,6 +149,31 @@ const utils = {
...
+
+      if (Array.isArray(value)) {
+        pair = utils.stringify(value, prefix ? `${prefix}[${encodedKey}]` : encodedKey, 'array');
+      } else if (typeof value === 'object') {
+        pair = utils.stringify(value, prefix ? `${prefix}[${encodedKey}]` : encodedKey, 'object');
+      } else if (objType === 'object') {
+        pair = `${prefix ? `${prefix}[${encodedKey}]` : encodedKey}=${utils.encodeURIComponentRFC3986(value)}`;
+      } else {
+        pair = `${prefix || encodedKey}=${utils.encodeURIComponentRFC3986(value)}`;
+      }
```
jjl014, line 170: Should we add a case to set `pair = ''` if the value is `undefined` or `null`? 🤔 I don't think it'll ever really happen where we'll have an undefined or null value, but might be good to safeguard against it.

Might be worth adding a test around this new function as well. This is what I tested with:

`utils.stringify({ a: '1', b: ['1,2'], c: ['2', '3'], d: [true, false], e: { f: ['g', 'h'] }, i: undefined, j: null })`

jjl014 (commented):
Awesome job! This is very cool! 🔥 

I left one comment around some weird edge cases and possibly adding a test around the new function, but otherwise lgtm!

Would also like to get another set of eyes on this since this affects quite a bit of stuff

src/utils/helpers.js
esezen: [4d29e01](https://github.com/Constructor-io/constructorio-client-javascript/pull/213/commits/4d29e0140712fa818eceec0b9863fcb04e13e80b)

jjl014 (approved):
Lgtm! 🚀

jjl014 (approved):
Looks good to me! 😄 🚀

sblaurock:
Please no ship before holiday period 😄

RohitProaxiom:
Looks good ! 😄 🚀

esezen:
Add test for emojis

PR #217, opened 2022-11-15
Lint fix + consistency adjustment.

sblaurock:
We don't do this parameters object operation in any of the other modules.

✅ All tests pass


esezen (approved):
LGTM!

PR #218, opened 2022-11-15
Optimize test runs in automated Actions environment

sblaurock:
- Conditionally run network timeout tests as they do not run reliably in Github actions environment
- Corrections to allow tests to be run in parallel (dependencies definitions missing in specific files)
- Add `:parallel` commands to allow tests to be run in parallel

Sync (171 seconds)
```
npm run test:all  35.18s user 2.64s system 22% cpu 2:51.89 total
```

Parallel (78 seconds)
```
npm run test:all:parallel  53.00s user 5.72s system 75% cpu 1:18.21 total
```

Parallel test run saves about 1 minute and 24 seconds when running the `all` tests job (standard + bundled)

We now run lint + testing on PR creation

Screenshot 2022-11-14 at 10 55 25 PM

View with [?w=1](https://github.com/Constructor-io/constructorio-client-javascript/pull/218/files?w=1)


esezen (approved):
🚢

PR #226, opened 2022-12-02
[NO-CSL] Update skipNetworkTimeoutTests

esezen:
When we renamed `RUN_NETWORK_TIMEOUT_TESTS` to `SKIP_NETWORK_TIMEOUT_TESTS` we forgot to negate the values. This PR doesn't change anything functionally, just updates the values to align with the variable name.


sblaurock (approved):
Thank you! Great catch.

PR #227, opened 2022-12-02
[NOCSL] Add support for item_name in browse click

stanlp1:
Customer does not expose item_ids so need to implement tracking based off of item_name only.


mocca102 (approved):
LGTM 👍

PR #231, opened 2022-12-22
No PSL - Add quiz service url

esevangelista:
- Add support for specifying service url for the quiz client
  - to enable quizzes for demos in both beta (Farmstand) and prod (prospects)
  - Related PR: https://github.com/Constructor-io/demo/pull/416


stanlp1 (approved):
Nice, this looks good to me. 

Two things.

1. Can we get rid of the dist file.

2. Can we add some test for the quizzes trailing slash.

Ex: https://github.com/Constructor-io/constructorio-client-javascript/blob/a5e6d0a8ddabff7d7263f2c2a39d83e10ab0fbed/spec/src/constructorio.js#L81-L90

esevangelista:
@sblaurock ready for deployment! this is not urgent so def no rush. Thanks!

sblaurock (changes requested):
Nice job @esevangelista!

Technically, the product is called `quizzes`, not `quiz`. To reduce confusion, it'd likely be best to update our variable naming.

Could we a check under the `Should return an instance with custom options when valid API key is provided` to verify that the quiz service URL is changed if a different one is specified? Thanks!

sblaurock:
One more thing - we should not have a change to the `dist` folder in PR. That will happen automatically when we bundle a new version.

sblaurock (approved):

PR #233, opened 2023-01-18
[CSL-1998] Add Quizzes tracking 

mocca102:
Postman collection https://constructor.slack.com/archives/C02RSEE9QE4/p1674071368260209


spec/src/utils/helpers.js
```diff
@@ -402,5 +404,62 @@ describe('ConstructorIO - Utils - Helpers', () => {
         expect(stringified).to.equal('a=1&b=1%2C2&c=2&c=3&d=true&d=false&e%5Bf%5D=g&e%5Bf%5D=h');
       });
     });
+    describe('toSnakeCase', () => {
+      it('Should convert to camel case', () => {
```
stanlp1, line 408: I think you mean 'Should convert to snake case' here.

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+
+      if (typeof url !== 'string') {
+        return new Error('"url" is a required parameter of type string');
+      }
+
+      bodyParams.quizId = quizId;
+      bodyParams.quizVersionId = quizVersionId;
+      bodyParams.url = url;
+
+      if (!helpers.isNil(section)) {
```
stanlp1, line 1282: Can we default to 'Products' for section?

I think we default to 'Products' for a lot of the other endpoints, the comments above also indicate default of product.

* @param {string} [parameters.section='Products'] - Index section

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+        if (typeof section !== 'string') {
+          return new Error('"section" must be a string');
+        }
+        queryParams.section = section;
+        bodyParams.section = section;
+      }
+
+      if (!helpers.isNil(resultCount)) {
```
stanlp1, line 1290: Do we need both isNil and typeof check?

Since typeof null/undefined won't be string/number so I don't think we need the nil check.

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+      }
+
+      if (typeof itemId !== 'string' && typeof itemName !== 'string') {
+        return new Error('"itemId" or "itemName" is a required parameter of type string');
+      }
+
+      bodyParams.quizId = quizId;
+      bodyParams.quizVersionId = quizVersionId;
+
+      if (!helpers.isNil(itemId)) {
```
stanlp1, line 1400: Same as above, do we need isNil?

Also, I think we can get rid of the itemId check like you did for quizId, versionId since we already check it above.

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+      }
+
+      if (!helpers.isNil(itemName)) {
+        if (typeof itemName !== 'string') {
+          return new Error('"itemName" must be a string');
+        }
+        bodyParams.itemName = itemName;
+      }
+
+      if (!helpers.isNil(section)) {
```
stanlp1, line 1414: Same as previous, should we default section to Produdcts?

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+      }
+
+      if (typeof itemId !== 'string' && typeof itemName !== 'string') {
+        return new Error('"itemId" or "itemName" is a required parameter of type string');
+      }
+
+      bodyParams.quizId = quizId;
+      bodyParams.quizVersionId = quizVersionId;
+
+      if (!helpers.isNil(itemId)) {
```
stanlp1, line 1548: Same as previous, I think we can get rid of this check.

stanlp1 (commented):
Left some nitpicky comments about just the code. Haven't had a chance to test it yet.

stanlp1 (approved):
LGTM, relevant tests pass. Lets have someone else take a pass too.

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+  }
+
+  /**
+   * Send quiz result click event to API
+   *
+   * @function trackQuizResultClick
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.quizId - Quiz Id
+   * @param {string} parameters.quizVersionId - Quiz version Id
+   * @param {string} [parameters.itemId] - Product item unique identifier
```
esezen, line 1341: I see you added logic to check one of itemId or itemName exists. Can we also add a note here saying you need to pass one of these?

src/modules/tracker.js
```diff
@@ -1228,6 +1229,401 @@ class Tracker {
...
+  }
+
+  /**
+   * Send quiz conversion event to API
+   *
+   * @function trackQuizConversion
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.quizId - Quiz Id
+   * @param {string} parameters.quizVersionId - Quiz version Id
+   * @param {string} [parameters.itemId] - Product item unique identifier
```
esezen, line 1486: Same as above

esezen (approved):
Left small comments. LGTM!

src/utils/helpers.js
```diff
@@ -182,6 +182,19 @@ const utils = {
 
     return allValues.join('&');
   },
+
+  toSnakeCase: (camelCasedStr) => camelCasedStr.replace(/[A-Z]/g, (prefix) => `_${prefix.toLowerCase()}`),
+
+  toSnakeCaseKeys: (camelCasedObj, toRecurse = false) => {
+    const snakeCasedObj = {};
```
esezen, line 189: Can we add some whitespace around the code blocks to separate them?

sblaurock:
One thing I feel we may want to address for consistency - all the parameters here are camelCase when the parameters in other tracking methods are snake_case. I know the snake_case part doesn't match the rest of the repository, but it may further add to the confusion if we suddenly change it now. What do you think @Constructor-io/customer-integrations ?

stanlp1:
> One thing I feel we may want to address for consistency - all the parameters here are camelCase when the parameters in other tracking methods are snake_case. I know the snake_case part doesn't match the rest of the repository, but it may further add to the confusion if we suddenly change it now. What do you think @Constructor-io/customer-integrations ?

https://github.com/Constructor-io/constructorio-node/pull/135

We have another story that I've been forgetting about that is doing this for all of the parameters to make it consistent everywhere.

esezen:
@sblaurock agreed. We have another story Stanley mentioned that would change everything to camelCase. We talked about this a couple of times with the team. I believe the only open item was to use something similar to lodash's implementation and use the nesting logic from our stringify logic. @jjl014 does that sound accurate?

jjl014:
@Constructor-io/customer-integrations I think @sblaurock has a good point on the snake vs camel casing here. There will be some discrepancies if we merge this PR. Should we change the parameters to be snake cased for now and migrate everything to be camel cased after in a separate PR to keep things consistent?

We have a PR in the [node repo for this](https://github.com/Constructor-io/constructorio-node/pull/139), but it doesn't look like we have an equivalent PR in the JS client.

jjl014:
@mocca102 Let's change everything to be camel cased for now and rip out the code around transforming things to the snake cased and create a separate PR for handling that (in addition to migrating all of the other tracking functions to be camel cased).

src/modules/tracker.js
```diff
@@ -1228,6 +1228,398 @@ class Tracker {
...
+   * Send quiz conversion event to API
+   *
+   * @function trackQuizConversion
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.quiz_id - Quiz Id
+   * @param {string} parameters.quiz_version_id - Quiz version Id
+   * @param {string} [parameters.item_id] - Product item unique identifier (Either item_id or item_name is required)
+   * @param {string} [parameters.item_name] - Product item name
+   * @param {string} [parameters.variation_id] - Product item variation unique identifier
+   * @param {string} [parameters.revenue] - The subtotal (not including taxes, shipping, etc.) of the entire order
```
esezen, line 1486: I think this line uses the description of trackPurchase but it should be using the description of trackConversion since this page would have individual items not an order with multiple items

src/modules/tracker.js
```diff
@@ -1228,6 +1228,398 @@ class Tracker {
...
+   *         url: www.example.com
+   *         result_count: 167,
+   *     },
+   * );
+   */
+  trackQuizResultsLoaded(parameters, networkParameters = {}) {
+    // Ensure parameters are provided (required)
+    if (parameters && typeof parameters === 'object' && !Array.isArray(parameters)) {
+      const requestPath = `${this.options.serviceUrl}/v2/behavioral_action/quiz_result_load?`;
+      const { quiz_id, quiz_version_id, url, section = 'Products', result_count, result_id, result_page } = parameters;
```
jjl014, line 1261: Could we break this into multiple lines like we did in the other functions?

jjl014 (approved):
Nice, this looks great to me @mocca102! 🚀  Thanks for working on this!

Aside from the comment that @esezen mentioned on the parameter description, we'll also want to take into account a new [quiz_session_id](https://github.com/Constructor-io/autocomplete/pull/4527) parameter that the Quizzes team is planning to add to the behavioral endpoints. Looks like it will be a required parameter.

We'll either want to coordinate with them and add that as an additional parameter to this PR or create a separate PR as a fast follow when they release the new parameter.

cc @Constructor-io/customer-integrations

src/modules/tracker.js
mocca102: Good catch

jjl014 (approved):
Awesome stuffs! Thank you for making all the changes, this is looking great to me!

I did make some adjustments to the descriptions of the parameters to match what we do in other places/functions. Hope that's okay!

Other than that, I think we should be good to go. @esezen would you mind taking a final look and merge/deploy if it looks good to you? Thanks! ❤️

PR #235, opened 2023-01-24
[CSL-2014] Update requests to point to the new index

esezen:


jjl014 (approved):
Very nice. All tests pass. Looking mighty fine to me! 🚀

PR #237, opened 2023-02-01
update verify node script and run npm install with node lts

mocca102:


Mudaafi (approved):
LGTM!

package.json
```diff
@@ -58,7 +59,7 @@
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-plugin-import": "^2.25.2",
     "husky": "^7.0.4",
-    "jsdoc": "^3.6.7",
+    "jsdoc": "^4.0.0",
```
esezen, line 62: I'm a bit scared this might break things. V4 is new and they don't have documentation for it. Is there a problem between jsdoc 3 and Node 18? It seemed to be working fine for me 🤔

esezen: Update: I am getting errors now 🤔

esezen:
Can we also update the GH workflows to make sure they use the same Node version we do?

PR #240, opened 2023-02-07
[CSL-1792] - Add voice search

mocca102:
API Documentation
https://docs.constructor.io/rest_api/search/natural_language_search/

Removed parameters because of no API support
- filters
- sort_order
- sort_by

Missing tests
- redirect_rule =>  if supported.

### DIFF
--- a/spec/src/modules/search.js
+++ b/spec/src/modules/search.js
@@ -981,8 +981,8 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
         expect(res).to.have.property('request').to.be.an('object');
         expect(res).to.have.property('response').to.be.an('object');
         expect(res).to.have.property('result_id').to.be.an('string');
-        expect(res.request.term).to.equal('peanut [ ]');
-        expect(requestUrl).to.include(encodeURIComponent(`peanut ${specialCharacters}`));
+        expect(res.request.term).to.equal(`${term} [ ]`);
+        expect(requestUrl).to.include(encodeURIComponent(`${term} ${specialCharacters}`));
         done();
       });
     });
--- a/spec/src/modules/search.js
+++ b/spec/src/modules/search.js
@@ -853,13 +853,10 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
         section,
         sortBy,
       }).then((res) => {
-        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
-
         expect(res).to.have.property('request').to.be.an('object');
         expect(res).to.have.property('response').to.be.an('object');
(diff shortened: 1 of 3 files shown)


src/modules/search.js
```diff
@@ -229,6 +231,87 @@ class Search {
...
+    let signal;
+
+    if (typeof AbortController === 'function') {
+      const controller = new AbortController();
+
+      signal = controller && controller.signal;
+
+      // Handle network timeout if specified
+      helpers.applyNetworkTimeout(this.options, networkParameters, controller);
+
```
Mudaafi, line 271: Could we delete this extra newline before the end of the block :3

src/modules/search.js
```diff
@@ -229,6 +231,87 @@ class Search {
...
   }
+
+  /**
+   * Retrieve voice search results from API
+   *
+   * @function getVoiceSearchResults
+   * @description Retrieve voice search results from Constructor.io API
+   * @param {string} query - Term to use to perform a voice search
+   * @param {object} [parameters] - Additional parameters to refine result set
+   * @param {number} [parameters.page] - The page number of the results(Can't be used together with offset)
```
Mudaafi, line 242: Could we add a space before the parentheses?

spec/src/modules/search.js
```diff
@@ -656,4 +656,477 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
...
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+
+        expect(res).to.have.property('request').to.be.an('object');
+        expect(res).to.have.property('response').to.be.an('object');
+        expect(res).to.have.property('result_id').to.be.an('string');
+        expect(res.request.sort_by).to.equal(sortBy);
+        expect(requestedUrlParams).to.have.property('sort_by').to.equal(sortBy);
+        done();
+      });
+    });
```
Mudaafi, line 865: It might be good to know if/when passing in sort options become [disallowed](https://docs.constructor.io/rest_api/search/natural_language_search/#http-request) but since the behaviour is not intended to be supported to begin with, should we remove this test?

Mudaafi (approved):
Some nit-picky linting on whitespaces but it LGTM! All the tests pass on my machine :)

Note: The tests for types fail but it's a separate thing so I'll put up a story for that one.

esezen (changes requested):
Thing I mentioned on Slack

src/modules/search.js
mocca102: Eyes of an Eagle

esezen (approved):
LGTM! @jjl014 would you like to take a final look or am I good to merge&deploy?

jjl014:
Looking good to me! Feel free to merge+deploy @esezen!

PR #242, opened 2023-03-03
[CSL-2101] Add v2 search events

stanlp1:


src/modules/tracker.js
```diff
@@ -108,6 +127,33 @@ class Tracker {
...
+  trackInputFocusV2(user_input = '', networkParameters = {}) {
+    const baseUrl = 'https://ac.cnstrc.com/v2/behavioral_action/focus?';
+    const bodyParams = { user_input };
+
+    const requestMethod = 'POST';
+    const requestBody = applyParams(bodyParams, {
+      ...this.options,
+      requestMethod,
+    });
+    this.requests.queue(`${baseUrl}${applyParamsAsString({}, this.options)}`, requestMethod, requestBody, networkParameters);
```
esezen, line 151: Not specific to this function, more of a question about all V2 endpoints. Do we still have to duplicate params in body of the POST and in the query params?

src/modules/tracker.js
```diff
@@ -402,6 +702,102 @@ class Tracker {
...
+   * Send click through event to API
+   * @private
+   * @function trackSearchResultClickV2
+   * @param {string} term - Search results query term
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.item_name - Product item name
+   * @param {string} parameters.item_id - Product item unique identifier
+   * @param {string} [parameters.variation_id] - Product item variation unique identifier
+   * @param {string} [parameters.result_id] - Search result identifier (returned in response from Constructor)
+   * @param {string} [parameters.result_count] - Number of results in total
```
esezen, line 715: Should this and the next one be `number`?

src/modules/tracker.js
```diff
@@ -402,6 +702,102 @@ class Tracker {
...
+   * @param {string} [parameters.result_page] - Current page of results
+   * @param {string} [parameters.result_position_on_page] - Position of selected items on page
+   * @param {string} [parameters.num_results_per_page] - Number of results per page
+   * @param {object} [parameters.selected_filters] - Key - Value map of selected filters
+   * @param {string} [parameters.item_is_convertible] - Whether or not an item is available for a conversion
```
esezen, line 720: Should this be `boolean`?

src/modules/tracker.js
```diff
@@ -402,6 +702,102 @@ class Tracker {
...
+          result_position_on_page,
+          num_results_per_page,
+          selected_filters,
+          section,
+        } = parameters;
+        const bodyParams = {
+          item_name,
+          name,
+          item_id,
+          customer_id,
```
esezen, line 760: Should we handle converting `customer_id` to `item_id` here? Same with name

esezen (commented):
Thanks for working on these changes @stanlp1 ! Changes are looking great. I left a few comments here and there. Can we also have a sync on plans to replace existing functions with these while maintaining backwards compatibility?

src/modules/tracker.js
stanlp1: Don't think so. The docs have query params, json params separate and there is no overlap.

https://docs.constructor.io/rest_api/behavioral_logging/autocomplete/

src/modules/tracker.js
```diff
@@ -340,6 +538,108 @@ class Tracker {
...
+          selected_filters,
+          url,
+          section,
+        } = parameters;
+        const queryParams = {};
+        let transformedItems;
+
+        if (items && Array.isArray(items) && items.length !== 0) {
+          if (typeof items[0] === 'string' || typeof items[0] === 'number') {
+            transformedItems = items.map((itemId) => ({ item_id: String(itemId) }));
```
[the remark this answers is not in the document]
stanlp1: Yea missed that, nice catch.

src/modules/tracker.js
```diff
@@ -281,6 +423,71 @@ class Tracker {
     return new Error('term is a required parameter of type string');
   }
 
+  /**
+   * Send autocomplete search event to API
+   * @private
+   * @function trackSearchSubmitV2
+   * @param {string} term - Term of submitted autocomplete event
```
jjl014, line 430: Should we rename this to `search_term` to match the API spec? 🤔

src/modules/tracker.js
```diff
@@ -340,6 +547,109 @@ class Tracker {
     return new Error('term is a required parameter of type string');
   }
 
+  /**
+   * Send search results loaded v2 event to API
+   * @private
+   * @function trackSearchResultsLoadedV2
+   * @param {string} term - Search results query term
```
jjl014, line 554: Should we change this to be `search_term` to match the API spec?

src/modules/tracker.js
```diff
@@ -340,6 +547,109 @@ class Tracker {
...
+          if (typeof items[0] === 'string' || typeof items[0] === 'number') {
+            transformedItems = items.map((itemId) => ({ item_id: String(itemId) }));
+          }
+        }
+
+        if (section) {
+          queryParams.section = section;
+        }
+
+        const bodyParams = {
```
jjl014, line 615: I think we're missing the `section` in the `bodyParams`

src/modules/tracker.js
```diff
@@ -281,6 +423,71 @@ class Tracker {
...
+   */
+  trackSearchSubmitV2(term, parameters, networkParameters = {}) {
+    // Ensure term is provided (required)
+    if (term && typeof term === 'string') {
+      // Ensure parameters are provided (required)
+      if (parameters && typeof parameters === 'object' && !Array.isArray(parameters)) {
+        const baseUrl = `${this.behavioralV2Url}search_submit?`;
+        const { original_query, user_input = original_query, group_id, section } = parameters;
+        const queryParams = {};
+        const bodyParams = { user_input, search_term: term };
```
jjl014, line 455: I think we're missing `section` in the `bodyParams`

src/modules/tracker.js
```diff
@@ -203,6 +259,92 @@ class Tracker {
...
+          user_input = original_query,
+          original_section,
+          section = original_section,
+          tr,
+          group_id,
+          item_id,
+          variation_id,
+        } = parameters;
+        const queryParams = {};
+        const bodyParams = {
```
jjl014, line 308: I think we're missing `section` in `bodyParams`

src/modules/tracker.js
```diff
@@ -281,6 +423,71 @@ class Tracker {
...
+   *     },
+   * );
+   */
+  trackSearchSubmitV2(term, parameters, networkParameters = {}) {
+    // Ensure term is provided (required)
+    if (term && typeof term === 'string') {
+      // Ensure parameters are provided (required)
+      if (parameters && typeof parameters === 'object' && !Array.isArray(parameters)) {
+        const baseUrl = `${this.behavioralV2Url}search_submit?`;
+        const { original_query, user_input = original_query, group_id, section } = parameters;
```
jjl014, line 453: Should we break this up into multiple lines like we do in the other functions? Same for `bodyParams` below

jjl014 (commented):
Nice, this is looking pretty good to me! Great job on handling the backwards compatibility!

I left a couple comments. Lmk what you think!

src/modules/tracker.js
stanlp1: Same as previous

src/modules/tracker.js
stanlp1: Same as previous

PR #243, opened 2023-03-18
[CSL - 2154] Migrate types from Autocomplete UI to Client-Javascript

Mudaafi:
Types Migrated: `Item` (formerly ItemBase and Item), `Product` & `Search Suggestions`. Converted them to interfaces to utilize the `extends` keyword as it allows consumers to more easily define their own "Items" based on their own Index Sections.

Could probably drop the `Product` & `Search Suggestions` types since you'd likely want to extend them to define your own data model for the `data` attribute but they're general types that always exists regardless of the index so I included them in.

On the Autocomplete OS UI side of things, we'd do the following:
```
import { Product as BaseProduct } from '@constructor-io/constructorio-client-javascript'

export type ProductData = { ... }
export interface Product extends BaseProduct {
    data: ProductData
}
```

### Possible New Stories:
- Replace client-js' individual SABR modules implementation of `Items` with these types to enforce consistency.

### DIFF
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -132,10 +132,10 @@ export interface FacetOption extends Record<string, any> {
 }
 
 export interface Group extends BaseGroup, Record<string, any> {
-  count: number;
-  data: Record<string, any>;
-  parents: BaseGroup[];
-  children: Group[];
+  count?: number;
+  data?: Record<string, any>;
+  parents?: BaseGroup[];
+  children?: Group[];
 }
 
 export interface Collection extends Record<string, any> {
@@ -194,6 +194,7 @@ export interface ItemData extends Record<string, any> {
   id?: string;
   image_url?: string;
   group_ids?: string[];
+  groups?: Group[]
   facets?: Facet[];
 }
 
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -180,3 +180,44 @@ export type FilterExpressionRange = {
 };
 
 export type FilterExpressionRangeValue = ['-inf' | number, 'inf' | number];
+
+export type Item = Product | SearchSuggestion | ItemBase;
+
+export interface ItemBase extends Record<string, any> {
+  id?: string;
+  url?: string;
+  value?: string;
+  section: string;
+  data?: Record<string, any>;
+}
+
+export type Product = {
+  is_slotted: boolean;
+  labels: Record<string, unknown>;
+  matched_terms: string[];
+  value: string;
+  data: {
+    facets: { name: string; values: string[] }[];
+    group_ids: string[];
+    id: string;
+    image_url: string;
+    price: number;
+    swatchColor: string;
+    url: string;
+    variation_id: string;
+  };
+  section: 'Products';
+};
+
+export type SearchSuggestion = {
+  is_slotted: boolean;
+  labels: Record<string, unknown>;
+  matched_terms: string[];
+  value: string;
+  data: {
+    id: string;
+    url?: string;
+  };
+  section: 'Search Suggestions';
+  id: string;
+};
--- a/src/types/index.d.ts
+++ b/src/types/index.d.ts
@@ -181,43 +181,19 @@ export type FilterExpressionRange = {
 
 export type FilterExpressionRangeValue = ['-inf' | number, 'inf' | number];
 
-export type Item = Product | SearchSuggestion | ItemBase;
-
-export interface ItemBase extends Record<string, any> {
-  id?: string;
-  url?: string;
-  value?: string;
-  section: string;
+export interface Item extends Record<string, any> {
+  value: string;
+  is_slotted: boolean;
+  labels: Record<string, unknown>;
+  matched_terms: string[];
   data?: Record<string, any>;
+  section?: string;
 }
 
-export type Product = {
-  is_slotted: boolean;
-  labels: Record<string, unknown>;
-  matched_terms: string[];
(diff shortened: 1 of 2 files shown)


src/types/index.d.ts
```diff
@@ -180,3 +180,20 @@ export type FilterExpressionRange = {
...
+export interface Item extends Record<string, any> {
+  value: string;
+  is_slotted: boolean;
+  labels: Record<string, unknown>;
+  matched_terms: string[];
+  data?: Record<string, any>;
+  section?: string;
+}
+
+export interface Product extends Item {
```
esezen, line 193: Autocomplete UI goes into more detail about the type. Should we do something similar here or do you think this level of specificity is enough?

```js
export type Product = {
  data: {
    facets: { name: string; values: string[] }[];
    group_ids: string[];
    id: string;
    image_url: string;
    url: string;
  };
  is_slotted: boolean;
  labels: Record<string, unknown>;
  matched_terms: string[];
  value: string;
  section: 'Products';
};

export type SearchSuggestion = {
  data: {
    id: string;
    url?: string;
  };
  id: string;
  is_slotted: boolean;
  labels: Record<string, unknown>;
  matched_terms: string[];
  value: string;
  section: 'Search Suggestions';
};
```

deomsj: yes please, for example: `image_url` exists on Products by not on SearchSuggestions

Mudaafi: Interesting. Just so we're on the same page, by default we would have these [optional, default parameters](https://docs.constructor.io/rest_api/autocomplete_queries/#products-common-fields) for `data` and if the customer has additional metadata they'd simply extend over it? 

It would probably make sense to have a base data class too then so you could do something like:

```
// In consumer code
export interface MySpecificDataObj extends BaseData {
    stars: number
}
export interface Product extends BaseProduct {
    data: MySpecificDataObj
}
```

Mudaafi: Made the recommended changes but also, @mocca102 and I had a pretty interesting discussion so I'm just going to record it here: 

With this PR, you'll be able to use your own version of `Products` or `Items` when you extend these types within your codebase and these types will be compatible when passed to Client-JS' functions.

However, when receiving these types back from our server, they'll default back to the default `Products` and `Items` types defined here. One solution to this would be to modify the existing function type definitions to be generic which allows consumers of this library to modify the types Constructor will be returning. This is a pretty nifty pattern that many TypeScript libraries follow, but we both agreed that the value doesn't justify the cost of implementing such a pattern. Still pretty interesting though.

src/types/index.d.ts
```diff
@@ -180,3 +180,31 @@ export type FilterExpressionRange = {
...
+  section: 'Products';
+}
+
+export interface SearchSuggestion extends Item {
+  data?: {
+    total_num_results?: number
+  } & ItemData;
+  section: 'Search Suggestions';
```
stanlp1, line 209: Why do we have a section field? 🤔

Mudaafi: oooh ur right, that's a good catch. I think in the Autocomplete Ui Library they needed to know, per-item, what kind of item it was (i.e. what section it belongs to).

this implementation has `section` as optional so it should be fine? but it might be confusing to customers using the client-js library :/ Do you think we should remove it?

src/types/index.d.ts
deomsj: agreed, the idea of applying this generics pattern here is very interesting. thanks for explaining this here, @Mudaafi . can imagine a world where we'll want to do this down the road, but agree with the decision to wait until that is truely needed and adding real value for real customer needs

src/types/index.d.ts
deomsj: ya 100% agreed. I think that removing `section` from the type here in our client-js library then extending this type over in our constructorio-ui-autocomplete repo makes the most sense :)

Mudaafi: I'll make the necessary changes, thanks for looking through this :)

stanlp1 (approved):
Types are looking better and better every day!.

src/types/index.d.ts
```diff
@@ -180,3 +180,28 @@ export type FilterExpressionRange = {
...
+  labels: Record<string, unknown>;
+  matched_terms: string[];
+  data?: ItemData;
+}
+
+export interface ItemData extends Record<string, any> {
+  url?: string;
+  id?: string;
+  image_url?: string;
+  group_ids?: string[];
```
esezen, line 196: Can we add `groups` here as well?

src/types/index.d.ts
```diff
@@ -132,10 +132,10 @@ export interface FacetOption extends Record<string, any> {
...
 
 export interface Group extends BaseGroup, Record<string, any> {
-  count: number;
-  data: Record<string, any>;
-  parents: BaseGroup[];
-  children: Group[];
+  count?: number;
+  data?: Record<string, any>;
+  parents?: BaseGroup[];
+  children?: Group[];
```
Mudaafi, line 138: Modified to follow the [Docs on Groups](https://docs.constructor.io/rest_api/search/groups_structure/) more accurately. I checked w/ Islam and this change doesn't affect any existing usage of this type as they were meant to be wrapped in `Partial<>` when used.

src/types/index.d.ts
Mudaafi: Added groups as an optional parameter as discussed

esezen (approved):
LGTM!

PR #248, opened 2023-04-11
[NO-CSL] Update qs docs

esezen:


jjl014 (approved):
LGTM! I think we might need to update the types as well in `browse.d.ts`

jjl014 (approved):
LGTM! Thank you! ❤️

PR #260, opened 2023-05-17
Update Package to be Isomorphic (work on both client and Node)

bertybot:
Hey guys,

Bert from Ashley here. Not really fully intending this PR to be merged but, more spark conversation. Right now we are in the process of converting a lot of our pages to Sveltekit (if you are familiar with Next.js it is pretty similar to that) One of the features of Sveltekit is having page level functions that can run on both the server and the client.  The only down side to this is that it requires clients like this to work on both server and client. Ie isomorphic. There are ways around it but, it complicates things a lot and this package seems extremly close to working on both.

Now experimenting with this package in Sveltekit the only thing that seems to be blocking this package from working is the lines deleted in this PR to get the sessionId, and ClientId with ConstructorIOID. I find this weird as the ConstructorIOID seems to be already Isomorphic having conditions to generate these values on Node and the browser. Is there a reason the API client doesn't want to generate these values on the server? Information on why would be much appreciated. 

At the very least is there anyway I can run the client on node without having to use the specific node package? It just seems really close to working and having to use 2 clients would up the complexity a lot. 

Thanks for your input!

### DIFF
--- a/src/constructorio.js
+++ b/src/constructorio.js
@@ -90,20 +90,8 @@ class ConstructorIO {
     let client_id;
     const versionFromGlobal = typeof global !== 'undefined' && global.CLIENT_VERSION;
 
-    // Initialize ID session if DOM context is available
-    if (helpers.canUseDOM()) {
-      ({ session_id, client_id } = new ConstructorioID(idOptions || {}));
-    } else {
-      // Validate session ID is provided
-      if (!sessionId || typeof sessionId !== 'number') {
-        throw new Error('sessionId is a required user parameter of type number');
-      }
-
-      // Validate client ID is provided
-      if (!clientId || typeof clientId !== 'string') {
-        throw new Error('clientId is a required user parameter of type string');
-      }
-    }
+    // Initialize ID session
+    ({ session_id, client_id } = new ConstructorioID(idOptions || {}));
 
     this.options = {
       apiKey,


esezen:
Hey @bertybot, thanks for opening this pull requests!

For hybrid applications, our recommendation is to use both of our SDKs together. We have an [example Next JS](https://github.com/Constructor-io/constructorio-nextjs-example) repo that shows how that can be done. I’ve not used Sveltekit personally but I believe it works in a pretty similar manner.

Our client-js package depends on some features/attributes that only exist in browsers. For example, the issue you found with writing and reading cookies using document. ConstructorioID package not only creates these values but it also writes them to cookies. When you are not in a browser, this becomes an issue since you don’t have direct access to cookies. One thing to note here - this is intentional since we want clientId and sessionId to stay consistent across user’s journey. If both the beacon and the server create cookies, that could lead to personalization not being applied correctly.

Another issue with using client-js on the server side instead of our Node package is we wouldn’t be able to verify the origin of the request. If you look at our Node package, you can see all functions take a userParameters argument. This argument lets you pass additional information about the origin of the request. When you are using our client-js on the frontend, this extra information is handled by the package since we have access to the browser (localStorage + cookies + window).

I understand it would be ideal if you can use one package for both browser and server but that wouldn’t allow you to pass [additional required parameters](https://github.com/Constructor-io/constructorio-node/wiki/Additional-Information-For-Backend-Integrations) for requests that are coming from the server. So our recommendation would be the following:

- We have the beacon on your site. The beacon will create and write clientId and sessionId to cookies.
- When a user goes to a page, if the initial render is happening on the server, please use the Node package and pass clientId and sessionId with the requests. These values don’t need to be created on the server side and can be read from the cookies (that were created by the beacon).
- On subsequent requests that are coming from the browser, please use client-js package. The package will automatically look for clientId and sessionId and pass them along in the requests.

There is one edge case where a new customer lands on a page directly with a link (maybe from Google results) and the server receives the request before the beacon can instantiate the respective cookies. In this case, it’s okay to not pass these values for the initial request.

esezen:
Thanks for the response @bertybot ! I'll follow up on Slack to schedule a call to understand your setup better

esezen:
Hey @bertybot , I tried reaching out on Slack a couple of times but didn’t hear back from you. If you are available to discuss this, would you either message me on Slack or email me at [enes@constructor.io](mailto:enes@constructor.io) . We can get something scheduled.

esezen:
Based on discussions, we decided to move forward with a 2 library approach to ensure all the required parameters are passed in respective places.

PR #263, opened 2023-06-16
[CSL-2482] Add hidden fields to recommendations

stanlp1:


esezen (approved):
LGTM!

spec/src/modules/recommendations.js
```diff
@@ -215,6 +215,25 @@ describe(`ConstructorIO - Recommendations${bundledDescriptionSuffix}`, () => {
...
+      const hiddenFields = ['hiddenField1', 'hiddenField2'];
+      const { recommendations } = new ConstructorIO({
+        apiKey: testApiKey,
+        fetch: fetchSpy,
+      });
+
+      recommendations.getRecommendations(podId, { itemIds, hiddenFields }).then((res) => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+
+        expect(res).to.have.property('request').to.be.an('object');
```
esezen, line 228: We are not really checking response[x].data for `hiddenFields` like we do for other endpoints but I believe the recommendation requests are returning 0 results in most of the cases since this is a test index. I think it's okay as long as the `request` object in the response has the correct `hiddenFields`

mocca102 (approved):
LGTM!

PR #266, opened 2023-07-10
Don't label passing of items / itemIds as optional in result load events

sblaurock:


mocca102:
@sblaurock Can we please update the typescript types as well?

esezen (approved):
LGTM!

PR #270, opened 2023-07-27
Update tracker items array type

mocca102:


src/types/index.d.ts
```diff
@@ -218,3 +218,8 @@ export interface VariationsMap {
   },
   dtype: 'array' | 'object'
 }
+
+export interface ItemsTracked {
```
esezen, line 222: Please double check but I believe this also accepts `itemId` optionally. Can we add it here?

esezen (commented):
Left one small comment. This is looking good

esezen (approved):
LGTM!

PR #273, opened 2023-08-22
[CSL-2740] Add items to quiz result load

stanlp1:
Adds support for items param in quiz result load

### DIFF
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -1737,7 +1737,7 @@ class Tracker {
    * @param {string} parameters.quizVersionId - Quiz version identifier
    * @param {string} parameters.quizSessionId - Quiz session identifier associated with this conversion event
    * @param {string} parameters.url - Current page url
-   * @param {object[]} parameters.items - List of product item unique identifiers in search results listing
+   * @param {object[]} parameters.items - List of product item objects
    * @param {string} [parameters.section='Products'] - Index section
    * @param {number} [parameters.resultCount] - Total number of results
    * @param {number} [parameters.resultPage] - The page of the results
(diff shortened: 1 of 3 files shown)


mocca102 (changes requested):
Can we add items parameter to trackQuizResultsLoaded in types as well?

mocca102 (approved):
LGTM!

src/modules/tracker.js
```diff
@@ -1737,6 +1737,7 @@ class Tracker {
    * @param {string} parameters.quizVersionId - Quiz version identifier
    * @param {string} parameters.quizSessionId - Quiz session identifier associated with this conversion event
    * @param {string} parameters.url - Current page url
+   * @param {object[]} parameters.items - List of product item unique identifiers in search results listing
```
esezen, line 1740: Can we update this to `parameters.items - List of product item objects` to match recommendations and browse?

src/modules/tracker.js
```diff
@@ -1775,9 +1778,22 @@ class Tracker {
         resultId = result_id,
         result_page,
         resultPage = result_page,
+        items,
       } = parameters;
       const queryParams = {};
       const bodyParams = {};
+      let transformedItems;
```
esezen, line 1785: I believe this code is from search results where we had to support string[] and object[]. Since this is new, we don't need to support string[]. Can we copy the implementation from browse?
```
      if (items && Array.isArray(items)) {
        bodyParams.items = items.slice(0, 100).map((item) => helpers.toSnakeCaseKeys(item, false));
      }
```

esezen (changes requested):
Left some small comments

esezen (approved):
LGTM!

PR #275, opened 2023-09-12
Add alias to events package to ensure no collision with node events

mocca102:
Create alias for events


esezen:
Can we close this in favor of #277?

PR #276, opened 2023-09-12
Export all types from index.d.ts

mocca102:
Instead of importing types from this path `@constructor-io/constructorio-client-javascript/lib/types ` We will be able to import types from this path `@constructor-io/constructorio-client-javascript`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.35.18",
   "description": "Constructor.io JavaScript client",
   "main": "lib/constructorio.js",
-  "types": "lib/types/constructorio.d.ts",
+  "types": "lib/types/index.d.ts",
   "scripts": {
     "clean": "sudo rm -rf node_modules package-lock.json",
     "version": "npm run verify-node-version && npm run docs && git add ./docs/* && npm run bundle && git add -A ./dist",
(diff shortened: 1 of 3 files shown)


esezen:
Would this break anything for projects using `@constructor-io/constructorio-client-javascript/lib/types` at the moment or would it be supporting both `@constructor-io/constructorio-client-javascript/lib/types` and `@constructor-io/constructorio-client-javascript`?

mocca102:
> Would this break anything for projects using `@constructor-io/constructorio-client-javascript/lib/types` at the moment or would it be supporting both `@constructor-io/constructorio-client-javascript/lib/types` and `@constructor-io/constructorio-client-javascript`?

It should support both. This screenshot for autocomplete-os-ui linking to this PR 
Screenshot 2023-09-12 at 8 44 17 PM

mocca102:
@esezen Made it backward compatible

esezen (commented):
I tested this by creating a new project and using this PR to install the package. It appears like both old and new way work as expected. Let's get one more review from the team before we merge but this is looking great! We might want to do the same for Node

stanlp1 (approved):
Tested in a new project, also seems to work as expected.

PR #281, opened 2023-09-25
Fix Type for sessionID

bertybot:
The constructor requires sessionId to be of type `number` but, the type for it is `string` causing unnecessary TypeScript errors


jjl014 (changes requested):
Thanks for catching this, @bertybot!

Could we also update the type in the doc comment [here](https://github.com/Constructor-io/constructorio-client-javascript/blob/bd83b48c6ff3c2ccd56e8c5a0d9c0d90f1251e40/src/constructorio.js#L44) in the constructorio module? 

Thank you!

jjl014 (approved):
Sweet, thanks again for catching this and making the updates! LGTM! 🚀

PR #282, opened 2023-09-27
[CSL-2863] Proactive pii detection

stanlp1:
Since all of our tracking requests go through the request queue, added logic in the request queue to not add a request to queue if PII is detected. Only the path, query parameters are checked at the moment.

### DIFF
--- a/src/utils/request-queue.js
+++ b/src/utils/request-queue.js
@@ -2,6 +2,7 @@
 const store = require('./store');
 const HumanityCheck = require('./humanity-check');
 const helpers = require('./helpers');
+const { requestContainsPii } = require('./helpers');
 
 const storageKey = '_constructorio_requests';
 const requestTTL = 180000; // 3 minutes in milliseconds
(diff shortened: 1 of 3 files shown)


esezen (approved):
LGTM! Diffend isn't happy about some packages but they don't seem related to this PR. We'll handle that separately

esezen:
@jjl014 do you want to take another look or am I good to merge & deploy?

src/utils/request-queue.js
```diff
@@ -33,6 +34,10 @@ class RequestQueue {
     if (this.sendTrackingEvents && !this.humanity.isBot()) {
       const queue = RequestQueue.get();
 
+      if (requestContainsPii(url, body)) {
+        return;
+      }
```
jjl014, line 39: We could probably move this right before line 35 where we get the queue.

jjl014 (approved):
Left one super minor comment, but this looks great to me! Shippit! 🚀 🚢

src/utils/request-queue.js
esezen: 🧠

stanlp1: 👍

PR #288, opened 2023-11-04
No ref/add get results config endpoint

v-i-s-h-n-u-ps:


src/modules/quizzes.js
```diff
@@ -214,6 +214,57 @@ class Quizzes {
         throw new Error('getQuizResults response data is malformed');
       });
   }
+
+  /**
+   * Retrieves filter expression and recommendation URL from given answers
```
esezen, line 219: I believe this description needs to be updated

src/modules/quizzes.js
```diff
@@ -214,6 +214,57 @@ class Quizzes {
...
+   *
+   * @function getQuizResultsConfig
+   * @description Retrieve quiz results page configuration from Constructor.io API
+   * @param {string} quizId - The identifier of the quiz
+   * @param {string} parameters - Additional parameters to refine result set
```
esezen, line 224: Maybe we can drop the `to refine the result set` part and only leave `additional parameters`

esezen (approved):
LGTM!
Left 2 small comments about doc strings

src/modules/quizzes.js
v-i-s-h-n-u-ps: [ac91e30](https://github.com/Constructor-io/constructorio-client-javascript/pull/288/commits/ac91e30bd1cff82d022b8477d3df6b67f0a673ef)

src/modules/quizzes.js
v-i-s-h-n-u-ps: [ac91e30](https://github.com/Constructor-io/constructorio-client-javascript/pull/288/commits/ac91e30bd1cff82d022b8477d3df6b67f0a673ef)

esezen (changes requested):
@v-i-s-h-n-u-ps it looks like two of the `getQuizResultsConfig` tests are failing. Would you be able to take a look?

esezen (approved):
LGTM!

PR #289, opened 2023-11-07
[CSL-2921] Remove store2 dependency

esezen:


src/utils/store.js
```diff
@@ -1,8 +1,124 @@
...
+
+    if (i >= sessionStorageLength) {
+      const overflowIndex = i - sessionStorageLength;
+
+      return Object.keys(this.overflow)?.[overflowIndex];
+    }
+
+    return keyFromSessionStorage;
+  },
+  length() { return sessionStorage.length; },
```
HHHindawy, line 53: Should we account for `overflow` length here as well? 
Food for thought 🤔

jjl014 (commented):
This is pretty freaking sweet! Nicely done!

Left some minor comments, but this is looking good to me! We should try to install this locally in our autocomplete-ui and do some testing with our beacons.

Since this change would effectively touch all of our beacons, I wonder if it would make sense to support a config option so we can turn it on for a single customer. If everything looks good and there aren't any issues, we can remove it and switch over completely. Lmk if you think that sounds unnecessary though. 😄

src/utils/store.js
esezen: Good catch!

spec/src/utils/store.js
```diff
@@ -0,0 +1,371 @@
...
+      expect(testData1[0]).to.be.equal('1');
+      expect(testData2[0]).to.be.equal('2');
+      expect(testData3[0]).to.be.equal('3');
+      expect(testData4[0]).to.be.equal('4');
+
+      store.local.remove('testData3');
+
+      const valueFromLocal = store.local.get('testData3');
+
+      expect(valueFromLocal).to.be.equal(null);
```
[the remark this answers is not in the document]
esezen: [00205af](https://github.com/Constructor-io/constructorio-client-javascript/pull/289/commits/00205af3ba7eb725199e5574839b02486ec0a760)

spec/src/utils/store.js
```diff
@@ -0,0 +1,371 @@
...
+      expect(testData1[0]).to.be.equal('1');
+      expect(testData2[0]).to.be.equal('2');
+      expect(testData3[0]).to.be.equal('3');
+      expect(testData4[0]).to.be.equal('4');
+
+      store.session.remove('testData3');
+
+      const valueFromSession = store.session.get('testData3');
+
+      expect(valueFromSession).to.be.equal(null);
```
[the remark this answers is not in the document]
esezen: [00205af](https://github.com/Constructor-io/constructorio-client-javascript/pull/289/commits/00205af3ba7eb725199e5574839b02486ec0a760)

spec/src/utils/store.js
```diff
@@ -0,0 +1,371 @@
...
+const oldStore = require('../../../test/utils/store_old'); // eslint-disable-line import/extensions
+
+chai.use(chaiAsPromised);
+chai.use(sinonChai);
+dotenv.config();
+
+// TODO: Add tests for bundled
+// const bundled = process.env.BUNDLED === 'true';
+
+describe('ConstructorIO - Utils - Store', () => {
```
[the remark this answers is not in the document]
esezen: Everything is stringified and parsed back so as long as JSON.stringify() and JSON.parse() works, we should be good. Couldn't hurt to add tests though:
[b58277c](https://github.com/Constructor-io/constructorio-client-javascript/pull/289/commits/b58277ca84e75b3d747f54941418e8e40fb98862)
Let me know if these cover everything

src/utils/store.js
```diff
@@ -1,8 +1,124 @@
...
+
+    if (i >= sessionStorageLength) {
+      const overflowIndex = i - sessionStorageLength;
+
+      return Object.keys(this.overflow)?.[overflowIndex];
+    }
+
+    return keyFromSessionStorage;
+  },
+  length() { return sessionStorage.length + Object.keys(this.overflow || {}).length; },
```
jjl014, line 53: I believe `this.overflow` will always be an object and it's defaulted to empty object, so the `|| {}` is probably not necessary.

src/utils/store.js
```diff
@@ -1,8 +1,124 @@
...
+
+    if (i >= localStorageLength) {
+      const overflowIndex = i - localStorageLength;
+
+      return Object.keys(this.overflow)?.[overflowIndex];
+    }
+
+    return keyFromLocalStorage;
+  },
+  length() { return localStorage.length + Object.keys(this.overflow || {}).length; },
```
jjl014, line 112: Same as above `|| {}` is probably not necessary.

jjl014 (approved):
Nice work, this is looking great! Thanks for making the changes!

Next step would be deciding how we want to roll this change out.

jjl014 (approved):
Thanks for making the updates. This is looking awesome! 🎉 

I would love to get another pair of eyes on this just in case we missed anything.

src/utils/store.js
```diff
@@ -1,8 +1,149 @@
...
+    }
+
+    if (typeof sessionStorage === 'undefined') {
+      return null;
+    }
+
+    const valueFromSession = sessionStorage.getItem(key);
+
+    if (valueFromSession) {
+      return JSON.parse(valueFromSession);
```
mocca102, line 18: If `valueFromSession`is invalid JSON that couldn't be parsed this would through an error that's not handled right? If so can we wrap this in a try catch to handle exception

src/utils/store.js
```diff
@@ -1,8 +1,149 @@
...
+      return valueFromOverflow;
+    }
+
+    if (typeof localStorage === 'undefined') {
+      return null;
+    }
+    const valueFromLocal = localStorage.getItem(key);
+
+    if (valueFromLocal) {
+      return JSON.parse(valueFromLocal);
```
mocca102, line 89: Same here couldn't this be an invalid JSON and it would through an error

mocca102 (changes requested):
Small comment about possible exceptions handling

mocca102 (approved):

mocca102 (approved):
LGTM!

PR #296, opened 2023-11-30
[NO-CSL] Update compile command

esezen:
Update compile command to do cleanup before it runs. Similar to bundle


jjl014 (approved):
LGTM! 🚀

PR #302, opened 2024-01-25
[PSL-3160] Proactive PII Defense - Update email regex + tests

vbartonicek:
This updates the email regex to exclude any strings containing a valid email. Check updated tests for more info.


deomsj:
For current scope of impact from various libs that we manage, I found this visual from kibana (below) and [this integrationMethodMap in our int dash code](https://github.com/Constructor-io/integration-dashboard/blob/f72416bdca28733ed015f5d03e7c7d4dc60d070a/src/frontend/utils/constants.js#L4) to be helpful:

Screenshot 2024-01-30 at 6 27 32 PM

Our "Frontend Beacon / UI" [delegates network calls](https://github.com/Constructor-io/autocomplete-ui/blob/76e5d5999118f34bda2b7cc636382c1fb8b501d3/src/tracker_v3.js#L8) to this client-js lib, so I feel great about where we are adding this layer of security / defensive programming.

deomsj (approved):
Thank you for leading the way on this @vbartonicek. Changes to util and related specs make sense and work great, nice work!

esezen (approved):
LGTM!

PR #304, opened 2024-02-21
update quiz results to support fmt options

v-i-s-h-n-u-ps:


esezen:
@stanlp1 to add hiddenFields and tests

src/types/quizzes.d.ts
```diff
@@ -21,11 +21,23 @@ export interface QuizzesParameters {
   quizSessionId?: string;
 }
 
+export interface QuizResultsFmtOptions {
```
esezen, line 24: @stanlp1  Since we don't support facets and groups at the moment, what do you think about updating the type here to only have `hidden_fields` and `fields`? We can add the other ones later as we add support

esezen (approved):
LGTM!

PR #307, opened 2024-03-27
[CSL 3345] Add Analytics Tags to Tracker Function

VincentGalloro:
Add Analytics Tags to Tracker Function

Note: added parameters to trackInputFocusV2. Will check that this function is not being used before merging (as this is a breaking change)

### DIFF
--- a/spec/src/modules/tracker.js
+++ b/spec/src/modules/tracker.js
@@ -1337,7 +1337,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         // Request
         expect(fetchSpy).to.have.been.called;
         expect(requestParams).to.have.property('variation_id').to.deep.equal(optionalParameters.variationId);
-        expect(requestParams).to.have.property('analytics_tag').to.deep.equal(testAnalyticsTag);
+        expect(requestParams).to.have.property('analytics_tags').to.deep.equal(testAnalyticsTag);
 
         // Response
         expect(responseParams).to.have.property('method').to.equal('POST');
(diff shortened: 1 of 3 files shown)


src/modules/tracker.js
```diff
@@ -133,16 +133,21 @@ class Tracker {
...
    */
-  trackInputFocusV2(userInput = '', networkParameters = {}) {
+  trackInputFocusV2(userInput = '', parameters = {}, networkParameters = {}) {
     const baseUrl = `${this.behavioralV2Url}focus?`;
     const bodyParams = {};
+    const {
+      analyticsTags,
+    } = parameters;
     let networkParametersNew = networkParameters;
     let userInputNew = userInput;
```
esezen, line 153: We can get rid of the 
```
    if (typeof userInput === 'object') {
      networkParametersNew = userInput;
      userInputNew = '';
    }
```
Since this won't be swapped with the legacy implementation anymore.

spec/src/modules/tracker.js
```diff
@@ -24,6 +24,7 @@ const bundled = process.env.BUNDLED === 'true';
 const skipNetworkTimeoutTests = process.env.SKIP_NETWORK_TIMEOUT_TESTS === 'true';
 const bundledDescriptionSuffix = bundled ? ' - Bundled' : '';
 const timeoutRejectionMessage = bundled ? 'AbortError: Aborted' : 'AbortError: The user aborted a request.';
+const testAnalyticsTag = { param1: 'test', param2: 'test2' };
```
esezen, line 27: Github won't let me add a comment to the correct line but can we add the `analyticsTags` to the trackInpuFocusV2 event on line 400

esezen (changes requested):
Thanks for working on these! I left some comments. Can we also update the types like we did in the [Node PR](https://github.com/Constructor-io/constructorio-node/pull/177)?

src/modules/tracker.js
VincentGalloro: I'm not totally sure I understand what to change here but I've moved userInput into parameters, let me know if that's not what we want

esezen: I was talking about removing this piece:
image
(I see you already did that)

esezen: Can we make `userInput` a parameter of its own? We follow that pattern is other places like `trackSearchSubmit(term, parameters, networkParameters = {})`.

VincentGalloro: Got it, moved `userInput` back into parameters

esezen (approved):
LGTM!

PR #308, opened 2024-03-29
[CSL-2646] Added Item Tracked Purchase type

HFaifman:
Added new Item Tracked Purchase type that includes price and count properties.

### DIFF
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -1087,11 +1087,11 @@ class Tracker {
    * Send purchase event to API
    * @function trackPurchase
    * @param {object} parameters - Additional parameters to be sent with request
-   * @param {{itemId: string | undefined
-   * variationId: string | undefined
-   * itemName: string | undefined
-   * count: number | undefined
-   * price: number | undefined}[]} parameters.items - List of product item objects
+   * @param {Array.<{itemId: string | undefined,
+   * variationId: string | undefined,
+   * itemName: string | undefined,
+   * count: number | undefined,
+   * price: number | undefined}>} parameters.items - List of product item objects
    * @param {number} parameters.revenue - The subtotal (excluding taxes, shipping, etc.) of the entire order
    * @param {string} [parameters.orderId] - Unique order identifier
    * @param {string} [parameters.section="Products"] - Index section
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -1085,6 +1085,7 @@ class Tracker {
 
   /**
    * Send purchase event to API
+   *
    * @function trackPurchase
    * @param {object} parameters - Additional parameters to be sent with request
    * @param {Array.<{itemId: string | undefined,
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -1085,10 +1085,13 @@ class Tracker {
 
   /**
    * Send purchase event to API
-   *
    * @function trackPurchase
    * @param {object} parameters - Additional parameters to be sent with request
-   * @param {object[]} parameters.items - List of product item objects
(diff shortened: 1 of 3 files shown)


HFaifman:
The failed test doesn´t seem to be related to these changes.

src/types/index.d.ts
```diff
@@ -231,3 +231,8 @@ export interface ItemTracked {
   itemId?: string;
   variationId?: string;
 }
+
+export interface ItemTrackedPurchase extends ItemTracked {
+  count?: number;
+  price?: number;
```
mocca102, line 237: The attributes here were defined as required on the Node SDK. Is there any reason we're making them optional here? https://github.com/Constructor-io/constructorio-node/commit/4e73d415c225aab4698f1b0c10ea4f584490b245#diff-d91c72b0c4408f78487da8710045c189275b3aee8df371b01b5ad358056f0255R342

mocca102 (changes requested):

mocca102 (commented):
I see we're passing items as is to the API but it's not apparent what items contains apart from the types. Can we document the items attributes here so it's clear what items should look like in the docs as well as the types

https://github.com/Constructor-io/constructorio-client-javascript/blob/csl-2646-js-client-sdk-switch-to-using-count-for-purchase-events/src/modules/tracker.js#L1083

src/types/index.d.ts
HFaifman: Well, the ticket made a request to ensure retrocompatibility.  I figured this was the best way to achieve that since required parameters would break existing implementations.
Not sure how much of that is relevant for this particular client though, I´m happy to correct it if you think it´s better.

HFaifman:
> I see we're passing items as is to the API but it's not apparent what items contains apart from the types. Can we document the items attributes here so it's clear what items should look like in the docs as well as the types
> 
> https://github.com/Constructor-io/constructorio-client-javascript/blob/csl-2646-js-client-sdk-switch-to-using-count-for-purchase-events/src/modules/tracker.js#L1083

This makes a lot of sense. Yes, I´ll add it to this PR.

src/types/index.d.ts
HFaifman: After a discussion with Islam, we will make these properties optional in the Node repo as well. I´ll open a PR shortly.

src/modules/tracker.js
```diff
@@ -1085,10 +1085,13 @@ class Tracker {
...
    * Send purchase event to API
-   *
    * @function trackPurchase
    * @param {object} parameters - Additional parameters to be sent with request
-   * @param {object[]} parameters.items - List of product item objects
+   * @param {{itemId: string | undefined
+   * variationId: string | undefined
+   * itemName: string | undefined
+   * count: number | undefined
+   * price: number | undefined}[]} parameters.items - List of product item objects
```
mocca102, line 1094: This looks fine to me but wanted to take a second opinion @esezen

mocca102 (approved):
LGTM!

src/modules/tracker.js
Herfaifman: This could be a typedef in jsdoc but I think that might be overkill since it´s not going to be used outside of here

esezen: I think this breaks jsdoc 🤔 
Screenshot 2024-04-02 at 4 15 57 PM

esezen: If there is no easy way, IMO, it's okay if we only have the example and not document the type here

HFaifman: This is weird...It seems to be picked up on fine by the IDE?

PR #309, opened 2024-04-01
[CSL-3045] Added enforcement of https for service client url

HFaifman:
This PR enforces HTTPs for the Service Url, similarly to what was done for the Node client in this [PR](https://github.com/Constructor-io/constructorio-node/pull/193)

### DIFF
--- a/spec/src/constructorio.js
+++ b/spec/src/constructorio.js
@@ -84,18 +84,12 @@ describe(`ConstructorIO${bundledDescriptionSuffix}`, () => {
   it('Should return an instance with correct serviceUrl when a http serviceUrl is passed', () => {
     const clientId = 'client-id';
     const sessionId = 'session-id';
-    const serviceUrl = 'https://constructor.io';
-    const quizzesServiceUrl = 'http://quizzes.constructor.io';
-    const version = 'custom-version';
-    const networkParameters = { timeout: 5000 };
+    const serviceUrl = 'http://constructor.io';
     const instance = new ConstructorIO({
       apiKey: validApiKey,
       clientId,
       sessionId,
       serviceUrl,
-      quizzesServiceUrl,
-      version,
-      networkParameters,
     });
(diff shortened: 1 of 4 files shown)


spec/src/constructorio.js
```diff
@@ -81,6 +81,27 @@ describe(`ConstructorIO${bundledDescriptionSuffix}`, () => {
     expect(instance.options).to.have.property('networkParameters').to.equal(networkParameters);
   });
 
+  it('Should return an instance with correct serviceUrl when a http serviceUrl is passed', () => {
```
esezen, line 84: The description says `http serviceUrl is passed` but we are passing an `https` value on line 87. Can we update the test to have an `http` value there. 
Also it might be nice to drop the other parameters and make it more similar to [this test](https://github.com/Constructor-io/constructorio-node/pull/193/files#diff-eeecf5f88f630f2dbfd79b209470f3d4707cd0cb828861731b3f36887fb8f29cR56) so it's clear what's being tested here

esezen (changes requested):
Left one comment about the test. The rest is looking good!

spec/src/constructorio.js
HFaifman: should be taken care of, thanks for pointing it out.

esezen (approved):
LGTM!

xkr47:
Hi! Would you approve adding an option to disable this "force https" feature?

I have implemented a "fake Constructor" in our java backend so we can prototype changes to our catalog updates on our local machines without having to actually upload the catalog to a Constructor index. If not, we'll have keep using a fork of this repo, which would mean added maintenance work and delayed updates for the client library. Nevertheless it is net beneficial for us by letting us develop our Constructor experience in an agile manner.

xkr47:
Welp, here it is anyway.. #460
EDIT: it was merged

PR #310, opened 2024-04-03
[CSL 2734] Handle Non-Json responses

VincentGalloro:
Handle non-json responses from server. If the server responds with plaintext we want to return an error indicating that the server responded with plaintext. The current error just says "invalid JSON ..." which doesn't make it obvious to the user that the underlying error came from a bad server response


VincentGalloro:
The failed test is likely due to the new ASA bundled tests. Don't think it's related to this PR

stanlp1 (approved):
LGTM from what I can tell everything works properly.

esezen (changes requested):
We have a failing test in the bundled version tests `npm run test:bundled`. It looks like this PR is adding `regeneratorRuntime` to the global window object. This is okay if it's intentional. Let's figure out if it is and if it's expected, we can update the test

jjl014:
@esezen @VincentGalloro I updated the util function `converResponseToJson` to use Promises instead of async/await. Seems to resolve the issue around `regneratorRuntime` and passes all tests. 👍

stanlp1 (approved):
LGTM.

esezen (approved):

PR #311, opened 2024-04-06
[CSL 3407] Add support for pre-filter expression for Recommendations

jjl014:
### Updates:
* Add pre-filter support to Recommendations
* Add test

### DIFF
--- a/src/modules/recommendations.js
+++ b/src/modules/recommendations.js
@@ -1,4 +1,4 @@
-/* eslint-disable object-curly-newline, no-param-reassign */
+/* eslint-disable object-curly-newline, no-param-reassign, no-underscore-dangle */
 const EventDispatcher = require('../utils/event-dispatcher');
 const helpers = require('../utils/helpers');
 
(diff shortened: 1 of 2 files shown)


Herfaifman (approved):

HFaifman (approved):
Code-wise it looks great. Let me know if it needs to be manually tested as well and I´ll whip it up locally.

src/modules/recommendations.js
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable object-curly-newline, no-param-reassign */
+/* eslint-disable object-curly-newline, no-param-reassign, no-underscore-dangle */
```
esezen, line 1: Do we still need this `no-underscore-dangle`? I'm guessing you removed some code that needed this

esezen (approved):
LGTM! Left one comment about the lint rule override

src/modules/recommendations.js
jjl014: Thanks for catching this! I think I added it for something else and then removed it. Will be removing the rule override 👍

PR #315, opened 2024-04-25
[CI-3437] Allows Purchases with the same orderId to be tracked for different keys

Mudaafi:
Gap has a single-cart system for multiple brands. This means we need to fire multiple purchase events on the beacon with the same orderId but to different keys. The current implementation prevents that, so I extended it to make it more flexible.

```
// previously
hasOrderId(orderId)

// now
hasOrderId({ orderId, apiKey })
hasOrderId({ orderId }) // also works
```

### DIFF
--- a/src/utils/helpers.js
+++ b/src/utils/helpers.js
@@ -113,8 +113,11 @@ const utils = {
     return null;
   },
 
-  hasOrderIdRecord(orderId) {
-    const orderIdHash = CRC32.str(orderId.toString());
+  hasOrderIdRecord({ orderId, apiKey }) {
+    let orderPerKeyId = orderId;
+    if (apiKey) orderPerKeyId = `${apiKey}-${orderId}`;
+
+    const orderIdHash = CRC32.str(orderPerKeyId.toString());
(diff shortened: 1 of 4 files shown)


src/utils/helpers.js
```diff
@@ -113,8 +113,11 @@ const utils = {
     return null;
   },
 
-  hasOrderIdRecord(orderId) {
-    const orderIdHash = CRC32.str(orderId.toString());
+  hasOrderIdRecord({ orderId, apiKey }) {
```
VincentGalloro, line 116: Does it matter that this change isn't backwards compatible? I guess as long as this helper is only called from the one spot it should be ok.

VincentGalloro (approved):
LGTM

src/utils/helpers.js
Mudaafi: yeah, this is an internal function from what I can tell so I figure a cleaner implementation is better than looping in backwards compatibility but this is also a really old function. Let's get some of the grand sages of the past to give a second review in case I missed something.

esezen (approved):
LGTM!

PR #316, opened 2024-05-06
[NO-CI] Fix version issue

esezen:


deomsj (approved):
👏🏻 Nice work, @esezen 👏🏻

thank you for the quick fix!

deomsj (approved):
👏🏻 Nice work, @esezen 👏🏻

thank you for the quick fix!

Mudaafi (approved):
Tried on a local sandbox w/ the bundled version and it works as expected. I wonder if we have this problem in os-ui-ac as well 🤔

sblaurock:
Would it be easier to just pull the version directly from package.json? Do we need a separate file to export the version number?

PR #318, opened 2024-05-23
[CI-3220]  Add ASA tracking functions and tests

stanlp1:
Checking with DS on some parameter definitions, expect those to be updated.

### DIFF
--- a/spec/src/modules/tracker.js
+++ b/spec/src/modules/tracker.js
@@ -8029,7 +8029,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
       section: 'Products',
     };
 
-    it('Should respond with a valid response when term and required parameters are provided', (done) => {
+    it('Should respond with a valid response when intent and required parameters are provided', (done) => {
       const { tracker } = new ConstructorIO({
         apiKey: testApiKey,
         fetch: fetchSpy,
--- a/spec/src/modules/tracker.js
+++ b/spec/src/modules/tracker.js
@@ -26,7 +26,7 @@ const bundledDescriptionSuffix = bundled ? ' - Bundled' : '';
 const timeoutRejectionMessage = bundled ? 'AbortError: Aborted' : 'AbortError: The user aborted a request.';
 const testAnalyticsTag = { param1: 'test', param2: 'test2' };
 
-describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
+describe.only(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
   let fetchSpy = null;
   let cleanup;
   const jsdomOptions = { url: 'http://localhost.test/path/name?query=term&category=cat' };
@@ -8312,7 +8312,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
     });
   });
 
-  describe('trackAssistantLoadStarted', () => {
+  describe.only('trackAssistantResultLoadStarted', () => {
     const requiredParameters = { intent: 'Show me cookie recipes' };
     const optionalParameters = {
       section: 'Products',
@@ -8346,7 +8346,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
     });
 
     it('Should respond with a valid response when term, required parameters and segments are provided', (done) => {
@@ -8372,7 +8372,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
     });
 
     it('Should respond with a valid response when required parameters and userId are provided', (done) => {
@@ -8398,7 +8398,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
     });
 
     it('Should respond with a valid response when required parameters and testCells are provided', (done) => {
@@ -8424,7 +8424,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
     });
 
     it('Should respond with a valid response when required and optional parameters are provided', (done) => {
@@ -8449,19 +8449,19 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(Object.assign(requiredParameters, optionalParameters))).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(Object.assign(requiredParameters, optionalParameters))).to.equal(true);
     });
 
     it('Should throw an error when no parameters are provided', () => {
       const { tracker } = new ConstructorIO({ apiKey: testApiKey });
 
-      expect(tracker.trackAssistantLoadStarted()).to.be.an('error');
+      expect(tracker.trackAssistantResultLoadStarted()).to.be.an('error');
     });
 
     it('Should throw an error when invalid parameters are provided', () => {
       const { tracker } = new ConstructorIO({ apiKey: testApiKey });
 
-      expect(tracker.trackAssistantLoadStarted()).to.be.an('error');
+      expect(tracker.trackAssistantResultLoadStarted()).to.be.an('error');
     });
 
     it('Should send along origin_referrer query param if sendReferrerWithTrackingEvents is true', (done) => {
@@ -8486,7 +8486,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
     });
 
     it('Should not send along origin_referrer query param if sendReferrerWithTrackingEvents is false', (done) => {
@@ -8511,7 +8511,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         done();
       });
 
-      expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+      expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
     });
 
     if (!skipNetworkTimeoutTests) {
@@ -8526,7 +8526,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
           done();
         });
 
-        expect(tracker.trackAssistantLoadStarted(requiredParameters, { timeout: 10 })).to.equal(true);
+        expect(tracker.trackAssistantResultLoadStarted(requiredParameters, { timeout: 10 })).to.equal(true);
       });
 
       it('Should be rejected when global network request timeout is provided and reached', (done) => {
@@ -8543,7 +8543,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
           done();
         });
 
-        expect(tracker.trackAssistantLoadStarted(requiredParameters)).to.equal(true);
+        expect(tracker.trackAssistantResultLoadStarted(requiredParameters)).to.equal(true);
       });
     }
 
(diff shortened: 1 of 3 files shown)


src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
     return new Error('parameters are required of type object');
   }
 
+  /**
+   * Send ASA request submitted event
+   *
+   * @function trackAssistantSubmit
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.intent - Intent of of user request
```
esezen, line 2243: `Intent of of user request`

src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
...
+
+    this.requests.send();
+
+    return new Error('parameters are required of type object');
+  }
+
+  /**
+   * Send assistant results page load started
+   *
+   * @function trackAssistantLoadStarted
```
esezen, line 2299: Not a big deal and I don't have a very strong preference but I would prefer `trackAssistantResultLoadStarted`. I think the name is a bit confusing with the word `result` in it

src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
...
+    return new Error('parameters are required of type object');
+  }
+
+  /**
+   * Send assistant results page load finished
+   *
+   * @function trackAssistantLoadFinished
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.intent - Intent of of user request
+   * @param {string} parameters.searchResultCount - Number of search results loaded
```
esezen, line 2357: `number`

src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
...
+   * @param {string} [parameters.section] - The section name for the item Ex. "Products"
+   * @param {string} [parameters.intentResultId] - The intent result id from the ASA response
+   * @param {object} [networkParameters] - Parameters relevant to the network request
+   * @param {number} [networkParameters.timeout] - Request timeout (in milliseconds)
+   * @returns {(true|Error)}
+   * @description Assistant results page load begun (but has not necessarily loaded completely)
```
esezen, line 2363: Copy paste from load start. I believe this needs to change

src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
...
+
+    this.requests.send();
+
+    return new Error('parameters are required of type object');
+  }
+
+  /**
+   * Send assistant results page load finished
+   *
+   * @function trackAssistantLoadFinished
```
esezen, line 2354: Same as above, we might want to add `result` in the name if we do it for the start function

src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
     return new Error('parameters are required of type object');
   }
 
+  /**
+   * Send ASA request submitted event
+   *
+   * @function trackAssistantSubmit
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.intent - Intent of of user request
+   * @param {string} [parameters.section] - The section name for the item Ex. "Products"
```
esezen, line 2244: Why do we have `[parameters.section="Products"]` in some functions but `[parameters.section]` in others? Is it intentional?

src/modules/tracker.js
```diff
@@ -2235,6 +2235,391 @@ class Tracker {
...
+
+    this.requests.send();
+
+    return new Error('parameters are required of type object');
+  }
+
+  /**
+   * Send ASA search submitted event
+   *
+   * @function trackAssistantSearchSubmit
```
esezen, line 2550: What is the difference between this and `trackAssistantSubmit`?

esezen (changes requested):
Thanks for working on this! I left some comments about the docs. Let me know what you think

src/modules/tracker.js
stanlp1: I am confused.

You prefer trackAssistantResultLoadStarted. But you also metnion that the name is confusing with the word result in it 😅 .

So are you for or against having result in it 🤔

src/modules/tracker.js
stanlp1: AFAIK, trackAssistantSubmit is for the initial ASA request.

trackAssistantSearchSubmit is more akin to trackSearchSubmit, when a user clicks on an alternative term that is presented for a specific search result.

src/modules/tracker.js
esezen: I meant without but I typed with 🤦 Sorry, I think it's a bit confusing when we do not have the word result in it

src/modules/tracker.js
mocca102: Did you get an answer from DS? @stanlp1

src/types/tracker.d.ts
```diff
@@ -253,5 +253,70 @@ declare class Tracker {
...
+      section?: string;
+    },
+    networkParameters?: NetworkParameters
+  ): true | Error;
+
+  trackAssistantSearchSubmit(
+    parameters: {
+      intent: string,
+      searchTerm: string;
+      userInput: string;
```
mocca102, line 312: What's the difference between intent, searchTerm, and userInput? :/

src/types/tracker.d.ts
```diff
@@ -253,5 +253,70 @@ declare class Tracker {
...
+      searchResultId: string;
+      groupId?: string;
+      section?: string;
+      intentResultId?: string;
```
mocca102, line 316: Are we missing search_result_id here?

src/types/tracker.d.ts
```diff
@@ -253,5 +253,70 @@ declare class Tracker {
...
+    networkParameters?: NetworkParameters
+  ): true | Error;
+
+  trackAssistantSearchSubmit(
+    parameters: {
+      intent: string,
+      searchTerm: string;
+      userInput: string;
+      searchResultId: string;
+      groupId?: string;
```
mocca102, line 314: The slab docs mentions that this is inside filters object

spec/src/modules/tracker.js
```diff
@@ -8022,4 +8022,1794 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
+      });
+
+      tracker.on('success', (responseParams) => {
+        const requestParams = helpers.extractBodyParamsFromFetch(fetchSpy);
+
+        // Request
+        expect(fetchSpy).to.have.been.called;
+        expect(requestParams).to.have.property('section').to.equal(optionalParameters.section);
+        expect(requestParams).to.have.property('intent_result_id').to.equal(optionalParameters.intentResultId);
+        expect(requestParams.filters).to.have.property('group_id').to.equal(optionalParameters.groupId);
```
mocca102, line 9653: Based on slab docs group_id should be inside a filters object

src/modules/tracker.js
```diff
@@ -2235,6 +2235,390 @@ class Tracker {
...
+   * @param {object} parameters - Additional parameters to be sent with request
+   * @param {string} parameters.intent - Intent of user request
+   * @param {string} parameters.searchTerm - Term of submitted assistant search event
+   * @param {string} parameters.userInput - The current assistant search query
+   * @param {string} parameters.searchResultId - resultId of search result the clicked item belongs to
+   * @param {string} [parameters.section] - The section name for the item Ex. "Products"
+   * @param {string} [parameters.groupId] - Group identifier of selected item
```
mocca102, line 2557: Skab document mentions groupId is inside of filter object

mocca102: It says "selected item" but not sure if this is accurate. Should it say selected search term maybe?

src/modules/tracker.js
```diff
@@ -2235,6 +2235,390 @@ class Tracker {
...
+    if (parameters && typeof parameters === 'object' && !Array.isArray(parameters)) {
+      // Ensure parameters are provided (required)
+      const baseUrl = `${this.behavioralV2Url}assistant_search_submit?`;
+      const {
+        section,
+        intent,
+        searchTerm,
+        userInput,
+        searchResultId,
+        groupId,
```
mocca102, line 2584: Should this be filters if we moved groupId inside filters object?

src/modules/tracker.js
```diff
@@ -2235,6 +2235,390 @@ class Tracker {
...
+        section,
+        search_term: searchTerm,
+        user_input: userInput,
+        search_result_id: searchResultId,
+        intent_result_id: intentResultId,
+      };
+
+      if (groupId) {
+        bodyParams.filters = { group_id: groupId };
+      }
```
mocca102, line 2599: I see you have filters here. Are we only going to support groupId and no other filters??

spec/src/modules/tracker.js
```diff
@@ -8022,4 +8022,1794 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
     });
   });
+
+  describe('trackAssistantSubmit', () => {
+    const requiredParameters = { intent: 'Show me cookie recipes' };
+    const optionalParameters = {
+      section: 'Products',
+    };
+
+    it('Should respond with a valid response when term and required parameters are provided', (done) => {
```
mocca102, line 8032: Do we need to keep "term" here since there's no real term?

src/types/tracker.d.ts
```diff
@@ -253,5 +253,70 @@ declare class Tracker {
...
+      intentResultId?: string;
+    },
+    networkParameters?: NetworkParameters
+  ): true | Error;
+
+  trackAssistantResultClick(
+    parameters: {
+      intent: string;
+      searchResultId: string;
+      itemId?: string;
```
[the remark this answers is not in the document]
stanlp1: So It's optional because either itemId or itemName are required. I saw that we just marked both optional in other types so I stuck with the same.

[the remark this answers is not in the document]
esezen: Makes sense. Thanks for the clarification

src/types/tracker.d.ts
stanlp1: I kept this the same as trackSearchSubmit, I think it's fine since there is only one.

src/modules/tracker.js
stanlp1: I believe so. This event is similar to trackSearchSubmit so I kept the parameters the same across them.

crgee1 (approved):
Looks good from what I can see, good stuff

src/modules/tracker.js
```diff
@@ -2235,6 +2235,390 @@ class Tracker {
...
+   * @param {string} [parameters.intentResultId] - The intent result id from the ASA response
+   * @param {object} [networkParameters] - Parameters relevant to the network request
+   * @param {number} [networkParameters.timeout] - Request timeout (in milliseconds)
+   * @returns {(true|Error)}
+   * @description Assistant results page load begun (but has not necessarily loaded completely)
+   * @example
+   * constructorio.tracker.trackAssistantResultLoadStarted(
+   *     {
+   *         intent: 'show me a recipe for a cookie',
+   *     },
```
jjl014, line 2312: Should we include `intentResultId` in the tracking examples?

src/modules/tracker.js
```diff
@@ -2235,6 +2235,390 @@ class Tracker {
...
+      } = parameters;
+      const queryParams = {};
+      const bodyParams = {
+        intent,
+        section,
+      };
+
+      if (section) {
+        queryParams.section = section;
+      }
```
jjl014, line 2273: I noticed that this was the only tracking function that applied `section` to the query params. Is that expected?

src/modules/tracker.js
jjl014: I'm curious if `groupId` or `filters` would even be applicable in this context. It is part of the original `search_submit` schema, but groupId in the context of search submit doesn't make a whole lot of sense. 🤔

jjl014 (commented):
Nice work on this! 🔥 I left a couple comments here and there.

I don't think we needed to implement all 6 tracking calls. I believe only the ones prefixed with `[MVT]` in the [document](https://constructor.slab.com/posts/approved-rfc-ai-shopping-assistant-behavioral-tracking-api-k8yxojn5#hk0hp-mvt-asa-request-submitted) were the ones that should be implemented. I think the other 3 needed to be scoped out a bit more, especially since there may be some things that aren't being surfaced in the ASA result API response.

We should look into scheduling a meeting with Alex S to discuss the things we're uncertain about. 

Either way, this is awesome. Thank you for working on this!

jjl014:
@stanlp1 Following up from the ASA call we had w/ Alex, `user_input` and `filters` can be removed from the assistant search submit. Alex is making an update to the schema endpoint to remove them as they don't seem necessary. 👍

src/modules/tracker.js
stanlp1: Nope, removed it.

stanlp1:
> @stanlp1 Following up from the ASA call we had w/ Alex, `user_input` and `filters` can be removed from the assistant search submit. Alex is making an update to the schema endpoint to remove them as they don't seem necessary. 👍

Removed, tests fail right now because the endpoint pr isn't merged yet.

jjl014 (approved):
LGTM! 🚀 

Let's get a re-review from another engineer on the team as well just in case I missed anything.

I'll follow up with Alex S when they're expecting to merge the schema change PR.

spec/src/modules/tracker.js
```diff
@@ -26,7 +26,7 @@ const bundledDescriptionSuffix = bundled ? ' - Bundled' : '';
 const timeoutRejectionMessage = bundled ? 'AbortError: Aborted' : 'AbortError: The user aborted a request.';
 const testAnalyticsTag = { param1: 'test', param2: 'test2' };
 
-describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
+describe.only(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
```
esezen, line 29: Only

esezen (approved):
One `.only` rest looks good to me!

PR #320, opened 2024-06-05
Add fields type to FmtOptions

mocca102:


stanlp1 (changes requested):
@mocca102  Where is this returned?

I don't see fields under fmtOptions. I think it might be under the base request. Also, should it be optional.

stanlp1 (approved):
LGTM. But let's make it optional.

PR #329, opened 2024-08-29
[CSL-3151] Update search results load to v2 api

stanlp1:


mocca102 (commented):
We need to update the types in `tracker.d.ts` as well

src/modules/tracker.js
```diff
@@ -585,9 +585,9 @@ class Tracker {
   }
 
   /**
-   * Send search results loaded v2 event to API
+   * Send search results loaded  event to API
    * @private
```
mocca102, line 589: I think we need to remove @private

src/modules/tracker.js
```diff
@@ -640,18 +642,19 @@ class Tracker {
...
+          url = helpers.getWindowLocation()?.href || 'N/A',
           section,
         } = parameters;
         const queryParams = {};
         let transformedItems;
 
         if (items && Array.isArray(items) && items.length !== 0) {
-          transformedItems = items;
+          const trimmedItems = items.slice(0, 100);
+          transformedItems = trimmedItems;
```
mocca102, line 653: I think this line can be removed and it will make no change right? 🤔

src/types/tracker.d.ts
```diff
@@ -58,7 +58,7 @@ declare class Tracker {
     networkParameters?: NetworkParameters
   ): true | Error;
 
-  trackSearchResultsLoadedV2(
+  trackSearchResultsLoaded(
     term: string,
     parameters: {
       url: string;
```
jjl014, line 64: We'll probably need to update the types to include `analyticsTags` as well

src/modules/tracker.js
```diff
@@ -640,18 +642,19 @@ class Tracker {
           sortBy = sort_by,
           selected_filters,
           selectedFilters = selected_filters,
-          url,
+          url = helpers.getWindowLocation()?.href || 'N/A',
```
jjl014, line 645: (Out of scope) Should we update browse load + rec view with the same logic as well to keep things consistent?

src/modules/tracker.js
```diff
@@ -616,7 +616,7 @@ class Tracker {
    *     },
    * );
    */
-  trackSearchResultsLoadedV2(searchTerm, parameters, networkParameters = {}) {
+  trackSearchResultsLoaded(searchTerm, parameters, networkParameters = {}) {
```
jjl014, line 619: Let's make sure we're able to support `analytics_tags` as well

spec/src/modules/tracker.js
```diff
@@ -1957,10 +1957,10 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
   describe('trackSearchResultsLoaded', () => {
     const term = 'Cat in the Hat';
     const requiredParameters = { numResults: 1337 };
-    const optionalParameters = { itemIds: [1, 2, 3] };
+    const optionalParameters = { item_ids: [1, 2, 3] };
     const legacyParameters = {
       ...requiredParameters,
-      customerIds: [1, 2, 3],
+      customer_ids: [1, 2, 3],
```
jjl014, line 1963: Any reason why we changed these to be snake_cased? It should work for both camel + snake right? 🤔

spec/src/modules/tracker.js
```diff
@@ -2070,41 +2072,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
       });
 
       // eslint-disable-next-line max-len
-      expect(tracker.trackSearchResultsLoadedV2(term, { ...requiredParameters, ...v2Parameters })).to.equal(true);
-    });
-
-    it('Backwards Compatibility - Should respond with a valid response when term and snake cased parameters are provided', (done) => {
```
jjl014, line 2076: Why did we remove this test? 🤔

spec/src/modules/tracker.js
```diff
@@ -7954,7 +7958,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
     it('Should throw an error when providing no messageType parameter', () => {
       const { tracker } = new ConstructorIO({ apiKey: testApiKey });
 
-      expect(tracker.on(null, () => {})).to.be.an('error');
+      expect(tracker.on(null, () => { })).to.be.an('error');
```
jjl014, line 7961: Random white space change

jjl014 (commented):
Looking pretty good to me!

Just left a couple comments and questions. Lmk what you think!

spec/src/modules/tracker.js
stanlp1: Not sure why I renamed them to be camel.

src/modules/tracker.js
stanlp1: 👍 I can make another story.

src/modules/tracker.js
```diff
@@ -626,9 +626,11 @@ class Tracker {
           num_results,
           numResults = num_results,
           result_count,
-          resultCount = numResults || result_count,
-          customer_ids,
-          item_ids,
+          resultCount = numResults || result_count || 0,
```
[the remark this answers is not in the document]
stanlp1: Makes sense to me 👍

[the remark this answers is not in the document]
jjl014: Did we want to move forward in making the change mentioned above?

jjl014 (approved):
Nice! Thanks for working on this and making the updates!

Things are looking great from what I can tell. Just left a follow-up comment, but otherwise I think this is ready to be shipped 🔥 🚢

stanlp1:
@esezen

src/types/tracker.d.ts
```diff
@@ -83,15 +83,7 @@ declare class Tracker {
       sortOrder?: string;
       sortBy?: string;
       section?: string;
-    },
-    networkParameters?: NetworkParameters
-  ): true | Error;
-
-  trackSearchResultsLoaded(
```
esezen, line 90: GH doesn't let me add a comment on line 74. Should that be changed to trackSearchResultsLoaded from trackSearchResultsLoadedV2?

PR #333, opened 2024-09-05
[CI-3839] Allow setting sessionID using setClientOptions

HHHindawy:
This PR allows setting session ID using `setClientOptions` for DOM-less environments only


src/constructorio.js
```diff
@@ -30,6 +31,31 @@ const computePackageVersion = () => {
   return `${versionPrefix}${versionModifiers.join('-')}${versionModifiers.length ? '-' : ''}${packageVersion}`;
 };
 
+const setSessionId = (sessionId, options) => {
```
jjl014, line 34: I could be missing something, but is there a reason why we aren't using the identity module to handle setting the session id?

I am slightly concerned that we need to reproduce the logic that currently exists in our identity module. I feel like most of the logic that we're introducing here should already be handled by the identity module itself and if it doesn't have support for what we need, we should be making adjustments there instead of the JS client.

We have access to the identity module from the JS client and I believe we have access to it's properties (i.e. `cookie_name_session_id`, `local_name_session_id`, etc) and the functions to set the cookie (i.e. `set_cookie`).

We should schedule a quick discussion to chat through the current implementation and see if it's possible to utilize the identity module to handle this.

jjl014 (commented):
Thanks for working on this!

I left a couple comments. I think we should chat with the team about the implementation before moving forward. Lmk what you think!

src/constructorio.js
```diff
@@ -169,6 +192,11 @@ class ConstructorIO {
         this.options.testCells = testCells;
       }
 
+      if (sessionId) {
+        this.options.sessionId = sessionId;
```
jjl014, line 196: Not a huge fan of saving the `sessionId` to the options. We'll have two places the value will reside, but we should ideally be using the value from localStorage/cookies.

I'm guessing this was mostly used for testing, but if we add `set_session_id` (mentioned above) and `get_session_id` to the identity module, we shouldn't need to keep track of this in multiple places.

jjl014 (commented):
This is looking a lot better!

I have some additional follow-ups. Let me know what you think!

I would love to get some thoughts from the @Constructor-io/customer-integrations team as well.

src/constructorio.js
HHHindawy: But this is already in options, I'm just updating it with the new value 🤔
Am I missing something?

jjl014: Ah, I completely missed that somehow. Please ignore me! 😅

stanlp1 (approved):
LGTM!

jjl014 (approved):
Nice! This looks great to me! 🚀 

Thanks for making the adjustments!

PR #336, opened 2024-10-01
Uit 37 client js pass extra query params to autocomplete requests

deomsj:
New Functionality:
- update [client js lib's autocomple module](https://github.com/Constructor-io/constructorio-client-javascript/blob/master/src/modules/autocomplete.js) to support an optional `qsParam` when calling `getAutocompleteResults`. Doing so should  pass a stringified version of this `qsParam` along to REST API requests in expected format (named `qs` param when sent along to the server)

Context: 
- Our client JS library already supports passing a `qsParam` query param with requests for [search](https://constructor-io.github.io/constructorio-client-javascript/module-search.html#~getSearchResults) and [browse results](https://constructor-io.github.io/constructorio-client-javascript/module-browse.html#~getBrowseResults)
- The REST API endpoint already supports a `qs` query param with requests for autocomplete results

Test for new functionality:
image

All tests are still passing:
image

I expect that docs we be updated programmatically when version command is run after merging these changes.


esezen (approved):
LGTM!

PR #337, opened 2024-10-01
[NO-CI] Make fetch available on the window for tests

esezen:


deomsj (approved):
Nice. thanks for the investigation and quick fix!

image

spec/src/constructorio.js
```diff
@@ -29,6 +29,9 @@ describe(`ConstructorIO${bundledDescriptionSuffix}`, () => {
 
     if (bundled) {
       ConstructorIO = window.ConstructorioClient;
+
+      // Make global fetch available on the window
+      window.fetch = fetch;
```
jjl014, line 34: I think we might be able to move this to the `jsdom-global.js` utils file so we don't need to specify it in all these files.

jjl014 (approved):
LGTM! 🚀 

Thanks for working on the fix! I just left one comment which might help clean things up a bit. 👍

spec/src/constructorio.js
esezen: Fantastic idea! I forgot we have this file. Updating the code

PR #338, opened 2024-10-13
Update search events group id param docs

mocca102:
Before
* @param {string} [parameters.groupId] - Group identifier of selected item

After
* @param {string} [parameters.groupId] - Group identifier of the group to search within if a user elected to search in a group, i.e. "Pumpkin in Canned Goods"

Changed in

- trackAutocompleteSelectV2
- trackAutocompleteSelect
- trackSearchSubmitV2
- trackSearchSubmit

### DIFF
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
 /* eslint-disable object-curly-newline, no-underscore-dangle, camelcase, no-unneeded-ternary */
 const EventEmitter = require('../utils/events');
 const helpers = require('../utils/helpers');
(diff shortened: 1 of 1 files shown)


src/modules/tracker.js
```diff
@@ -277,7 +278,7 @@ class Tracker {
    * @param {string} [parameters.tr] - Trigger used to select the item (click, etc.)
    * @param {string} [parameters.itemId] - Item id of the selected item
    * @param {string} [parameters.variationId] - Variation id of the selected item
-   * @param {string} [parameters.groupId] - Group identifier of selected item
+   * @param {string} [parameters.groupId] - Group identifier of the group to search within if a user elected to search in a group, i.e. "Pumpkin in Canned Goods"
```
Mudaafi, line 281: Just a suggestion: Group identifier of the group to search within. Only required if searching within a group, i.e. "Pumpkin in Canned Goods"

Feels like we should emphasize its very optional

Mudaafi (approved):
left one comment but lgtm

src/modules/tracker.js
mocca102: https://github.com/Constructor-io/constructorio-client-javascript/pull/338/commits/a87406c6ace42cc3a0c12d27b5bfa7197d3be4f9

PR #341, opened 2024-10-16
[CI-3913] Check for status of local storage in storage helpers.

stanlp1:
Adds a function to check if local storage is accessible before we try to access it.

In browsers that have cookies entirely disabled, localStorage, sessionStorage also become inaccessible. Even accessing the variable will throw an exception "This operation is insecure".

This change ensures that we can use storage before we attempt to access it.

### DIFF
--- a/src/utils/store.js
+++ b/src/utils/store.js
@@ -57,7 +57,7 @@ const session = {
       delete this.overflow[key];
     }
 
-    if (!canUseStorage('sessionStorage') || typeof sessionStorage !== 'undefined') {
+    if (canUseStorage('sessionStorage') || typeof sessionStorage !== 'undefined') {
       sessionStorage.removeItem(key);
     }
   },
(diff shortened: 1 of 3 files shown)


src/utils/store.js
```diff
@@ -1,3 +1,22 @@
...
+  try {
+    storage = window[type];
+    const x = '__storage_test__';
+    storage.setItem(x, x);
+    storage.removeItem(x);
+    return true;
+  } catch (e) {
+    return (
+      e instanceof DOMException
+      && e.name === 'QuotaExceededError'
```
mocca102, line 12: Why do we make an exception for this error only?

mocca102 (approved):
LGTM!

src/utils/store.js
stanlp1: This was taken from here.

https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability

I am guessing that if the Quota is Exceeded, storage is still technically available. Just full.

PR #346, opened 2024-10-24
[CI 3890] switch beforeunload with visibilitychange

crgee1:
Loom showing visibilityChange event firing as expected

https://www.loom.com/share/f4e8b2c2c1a047c09596c3bc819480f6

### DIFF
--- a/src/utils/request-queue.js
+++ b/src/utils/request-queue.js
@@ -20,7 +20,7 @@ class RequestQueue {
       : false; // Defaults to 'false'
 
     // Mark if page environment is unloading
-    helpers.addEventListener('pagehide', () => {
+    helpers.addEventListener('visibilitychange', () => {
       this.pageUnloading = true;
     });
 
--- a/src/utils/request-queue.js
+++ b/src/utils/request-queue.js
@@ -21,7 +21,9 @@ class RequestQueue {
 
     // Mark if page environment is unloading
     helpers.addEventListener('visibilitychange', () => {
-      this.pageUnloading = true;
+      if (document.visibilityState === 'hidden') {
+        this.pageUnloading = true;
+      }
     });
 
     if (this.sendTrackingEvents) {
--- a/src/utils/request-queue.js
+++ b/src/utils/request-queue.js
@@ -19,14 +19,14 @@ class RequestQueue {
       ? true
       : false; // Defaults to 'false'
 
-      helpers.addEventListener('visibilitychange', () => {
+    helpers.addEventListener('visibilitychange', () => {
       // Mark if page environment is unloading
       if (document.visibilityState === 'hidden') {
         this.pageUnloading = true;
       }
 
       // Send events once page is visible
-      if (this.pageUnloading = true && document.visibilityState === 'visible') {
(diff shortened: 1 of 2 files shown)


src/utils/request-queue.js
```diff
@@ -182,7 +182,7 @@ class RequestQueue {
       if (this.options && this.options.trackingSendDelay === 0) {
         this.sendEvents();
       } else {
-        // Defer sending of events to give beforeunload time to register (avoids race condition)
```
mocca102, line 185: What happened/happens when beforeunload/visibilitychange registers?

src/utils/request-queue.js
```diff
@@ -20,7 +20,7 @@ class RequestQueue {
       : false; // Defaults to 'false'
 
     // Mark if page environment is unloading
-    helpers.addEventListener('beforeunload', () => {
+    helpers.addEventListener('visibilitychange', () => {
```
esezen, line 23: What do you think about checking the `visibilityState` here? I think we only want to do this when it's `hidden`

```
 if (document.visibilityState === 'hidden') {
   ...
}
```

src/utils/request-queue.js
crgee1: beforeunload/visibilitychange changes the `pageUnloading` value and if the page is unloading than we wouldn't send events and wait for the next page to send those events from the

esezen (approved):

src/utils/request-queue.js
```diff
@@ -20,8 +20,10 @@ class RequestQueue {
       : false; // Defaults to 'false'
 
     // Mark if page environment is unloading
-    helpers.addEventListener('beforeunload', () => {
-      this.pageUnloading = true;
+    helpers.addEventListener('visibilitychange', () => {
+      if (document.visibilityState === 'hidden') {
```
jjl014, line 24: `visibilityState=hidden` based on what [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) says can mean the  "document is either a background tab or part of a minimized window, or the OS screen lock is active".

Would it make sense to add in logic to add in some additional logic here to call `send()` if the page becomes visible again? 🤔 

```
if (document.visibilityState === 'visibile') {
    this.send();
}
```

jjl014 (commented):
Thanks for working on this change!

The changes make sense to me, but I also left a comment around the `visibilitychange` listener. Let me know what you think. Would love to @esezen's thought as well since he mentioned the hidden change previously.

It seems like the following tests are failing as well. 🤔 

Screenshot 2025-01-10 at 5 24 32 PM

src/utils/request-queue.js
esezen: Sorry I somehow completely missed the notification for this one. I don't see any reason not to. Good call

jjl014 (approved):
The changes look good from what I can tell. I am slightly hesitant about releasing this to all beacon customers at the same time though.

Have we tested this out in the autocomplete-ui repo yet? I believe we should be able to install this version of the library by pointing it to this specific branch (can do this locally as well).

If things are working fine and there are no issues with events firing, then we should probably be okay to move forward. We'll want to keep an eye out on the events after deploying the changes.

PR #356, opened 2024-11-20
[NO-CI] Update search results loaded event

esezen:


src/modules/tracker.js
```diff
@@ -657,21 +656,19 @@ class Tracker {
           sortBy = sort_by,
           selected_filters,
           selectedFilters = selected_filters,
-          url = helpers.getWindowLocation()?.href || 'N/A',
+          url,
           section,
           analyticsTags,
-          resultCount = numResults || result_count || items?.length || 0,
```
jjl014, line 663: Should we keep this change as well?

jjl014 (changes requested):
Lots of changes in this PR, but things are mostly looking good from what I can tell. I did leave some comments around a few failing tests and some things that we may want to keep around. Lmk what you think. Thanks!

It might also be a good idea to get another review of the changes just in case I missed anything. 👍

spec/src/modules/tracker.js
```diff
@@ -2029,10 +2029,10 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
       });
 
       // eslint-disable-next-line max-len
-      expect(tracker.trackSearchResultsLoaded(term, snakeCaseParameters)).to.equal(true);
+      expect(tracker.trackSearchResultsLoadedV2(term, snakeCaseParameters)).to.equal(true);
     });
 
-    it('Should respond with a valid response when term and required parameters are provided', (done) => {
+    it('V2 Should respond with a valid response when term and required parameters are provided', (done) => {
```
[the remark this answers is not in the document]
esezen: Fixed

src/modules/tracker.js
```diff
@@ -1428,6 +1500,7 @@ class Tracker {
       const {
         section = 'Products',
         result_count,
+        resultCount = result_count,
```
jjl014, line 1503:
```suggestion
        resultCount = result_count ?? items?.length ?? 0,
```

jjl014 (approved):
LGTM!

Just left one minor comment/suggestion, but I think this is ready to go. Thank you for working on this!

PR #362, opened 2024-12-17
[NOCSL] Migrate to v2 search result load

stanlp1:


spec/src/modules/tracker.js
```diff
@@ -2077,71 +2079,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
       });
 
       // eslint-disable-next-line max-len
-      expect(tracker.trackSearchResultsLoadedV2(term, { ...requiredParameters, ...v2Parameters })).to.equal(true);
-    });
-
-    it('Backwards Compatibility - Should respond with a valid response when term and snake cased parameters are provided', (done) => {
```
jjl014, line 2083: I think I might've commented on this in the old PR as well, but should we keep this test for snake cased params, around?

jjl014 (approved):
The changes are looking great from what I can tell!

Just left one minor comment on one of the tests being removed.

spec/src/modules/tracker.js
stanlp1: Yea you did 😂 . I think the test is still there. We just had separate ones for v1/v2 before so we deleted one of them.

PR #365, opened 2025-01-09
[NOCI] Bot Check Fix

jjl014:
### Updates:
* Switch to use dynamic helper instead of depending on the instance state for determining human-ness. The session storage variable should be the source of truth


mocca102 (approved):
This looks good to me but lets get a second review

esezen (approved):

PR #368, opened 2025-02-26
[NOCSL] Request Queue - handle domlesss case

stanlp1:
Handle case where user is domless (react-native) and tracking requests don't get sent.

### DIFF
--- a/src/utils/humanity-check.js
+++ b/src/utils/humanity-check.js
@@ -57,6 +57,11 @@ class HumanityCheck {
       return true;
     }
 
+    // Bypass Storage/Event check if DOM not available.
+    if (!helpers.canUseDOM()) {
+      return false;
+    }
+
     // If the user hasn't performed a human event, it indicates it is a bot
     if (!this.getIsHumanFromSessionStorage()) {
       return true;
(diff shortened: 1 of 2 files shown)


src/utils/humanity-check.js
```diff
@@ -57,6 +57,11 @@ class HumanityCheck {
       return true;
     }
 
+    // Bypass Storage/Event check if DOM not available.
+    if (!helpers.canUseDOM()) {
```
jjl014, line 61: Should we move this logic to the [queue](https://github.com/Constructor-io/constructorio-client-javascript/blob/80c917cd3f7dda7f0dad348fa371562d9d9450e9/src/utils/request-queue.js#L34) function since we also have this similar check for [sendEvents](https://github.com/Constructor-io/constructorio-client-javascript/blob/80c917cd3f7dda7f0dad348fa371562d9d9450e9/src/utils/request-queue.js#L57) to keep similar logic in the same place. For example:

`this.sendTrackingEvents && (!this.humanity.isBot() || !helpers.canUseDOM())`

Or maybe we can just remove the check in `sendEvents` and handle it here instead since it also utilizes the isBot() check.

What do you think?

iIt might be worth adding the comment from `sendEvents` as well so we know why we're doing it: `// Consider user "human" if no DOM context is available`

jjl014 (approved):
Looking good from what I can tell! Just left one comment.

src/utils/humanity-check.js
stanlp1: Makes sense 👍

PR #369, opened 2025-03-14
[CI-4284] Update AutocompleteResponse

esezen:


src/types/autocomplete.d.ts
```diff
@@ -6,6 +6,7 @@ import {
   RequestFeatureVariant,
   VariationsMap,
   FilterExpression,
+  FmtOptions,
```
HHHindawy, line 9: Thanks for fixing these!

HHHindawy (approved):

PR #371, opened 2025-03-24
[CI-4285] Update tests

esezen:


spec/src/modules/tracker.js
```diff
@@ -3550,7 +3550,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
 
         // Response
         expect(responseParams).to.have.property('method').to.equal('POST');
-        expect(responseParams).to.have.property('message').to.equal('conversion type must be one of add_to_wishlist, add_to_cart, like, message, make_offer, read. If you wish to use custom types, please set is_custom_type to true and specify a display_name.');
+        expect(responseParams).to.have.property('message').to.equal('Invalid parameters');
```
esezen, line 3553: The error message has been moved to `response.errors[x].message` but we don't emit that in our RequestQueue at the moment on errors. I didn't want to change the implementation just because of this

stanlp1 (approved):
@esezen  Looks like a couple other [tests](https://github.com/Constructor-io/constructorio-client-javascript/blob/a078ef139812413a40876ed2428f8345998fbb5a/spec/src/modules/tracker.js#L6319-L6350) are failing because we no longer validate item id when beacon=false.

So we aren't returning an error anymore. 

Changes here look good but I don't know if we want to wait for backend to finalize their changes, or include these changes in this PR.

spec/src/modules/tracker.js
jjl014: Do you know if this is how errors are going to be surfaced moving forward? If so, it shouldn't be too hard to add it as a part of the emitted event (and it should be backwards compatible since it's an addition). If it's something that's expected to continue changing, I wonder if we should just spread (`...`) the JSON body into the emitted event. For example:

```
instance.eventemitter.emit('error', {
    url: nextInQueue.url,
    method: nextInQueue.method,
    ...(json && {...json}) // bringing back of one of Stanley's signature existence-check and spread
});
```

jjl014 (approved):
Changes look good from what I can tell. Just left a minor comment/question.

Thanks for fixing them! ❤️ 

> Looks like Jimmy created another PR to fix the tests so NVM.
I am not aware of this 😂

spec/src/modules/tracker.js
esezen: I like that. Should I create a story for it?

PR #376, opened 2025-05-05
[CI-4396] Add optional price attribute to ItemTracked type

TarekAlQaddy:


Mudaafi (approved):
lgtm, thanks for fixing spellcheck as well

PR #377, opened 2025-05-30
[NOCI] TrackConversion - Remove Result ID from Docs

VincentGalloro:
Removing result ID from the docs since it isn't used in the track conversion call


VincentGalloro:
> I believe we wanted to remove it from trackConversion example.
> 
> Also we don't need to remove it from docs/ files these are automatically generated.
> 
> Lets not remove the parameter from recommendations tracking methods in case a customer uses them

Honestly I'm not even sure how I accidentally removed it from track Recs instead of conversions. I think my brain stopped working that day. I've fixed it now, thank you for catching that

mocca102 (approved):

PR #386, opened 2025-07-15
[CI-4659] Rename Assistant Module to Agent

evanyan13:
## PR Type

What kind of change does this PR introduce?
- [X] Refactoring (no functional changes, no API changes)

## ✅ Definition of done:

* `agent` import works identically to current `assistant`
* `assistant` remains as a deprecated alias.

### Tasks
- [x] Move existing functions to `agent` module  
  - [x] Rename all `assistant` functions to `agent`

- [x] Import agent functions from `assistant` module  
  - [x] Export with backwards compatible names  
  - [x] Add JSDoc deprecation notice to all functions

- [x] Rename all `assistant` functions to `agent` in `tracker` module  
  - [x] Export with backwards compatible names  
  - [x] Add JSDoc deprecation notice to all functions

- [x] Add tests for `agent` module in `/spec`  
  - [x] Retain tests for `assistant` module to support backwards compatibility

- [x] Update README, docs, and TypeScript defs to reference `agent`

## Note:
* Can only pass `domain=assistant` as a query param to the endpoint: https://agent.cnstrc.com/v1/intent/ now.
  * No need to change `domain=assistant` – this is customer/integration-specific and may be any arbitrary value to distinguish multiple setups for the same ac_key.


src/constructorio.js
```diff
@@ -38,7 +39,8 @@ class ConstructorIO {
    * @param {string} parameters.apiKey - Constructor.io API key
    * @param {string} [parameters.serviceUrl='https://ac.cnstrc.com'] - API URL endpoint
    * @param {string} [parameters.quizzesServiceUrl='https://quizzes.cnstrc.com'] - Quizzes API URL endpoint
-   * @param {string} [parameters.assistantServiceUrl='https://assistant.cnstrc.com'] - AI Assistant API URL endpoint
+   * @param {string} [parameters.agentServiceUrl='https://agent.cnstrc.com'] - AI Shopping Agent API URL endpoint
+   * @param {string} [parameters.assistantServiceUrl='https://assistant.cnstrc.com'] - AI Shopping Assistant API URL endpoint
```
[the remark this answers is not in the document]
evanyan13: Thank you for noticing this. Have added the notice 👍

src/modules/agent.js
```diff
@@ -0,0 +1,187 @@
...
+  // eslint-disable-next-line no-param-reassign
+  eventSource.onerror = (error) => {
+    controller.error(error); // Pass the error to the stream
+    eventSource.close(); // Close the EventSource connection
+  };
+}
+
+/**
+ * Interface to agent SSE.
+ * Replaces the previous Assistant module.
```
mocca102, line 107: This will not mean much after we remove the Assistant module. What do you think about having a description for the Agent module itself like what it is

src/types/tracker.d.ts
```diff
@@ -266,6 +266,71 @@ declare class Tracker {
...
+      userInput: string;
+      searchResultId: string;
+      groupId?: string;
+      section?: string;
+      intentResultId?: string;
+    },
+    networkParameters?: NetworkParameters
+  ): true | Error;
+
   trackAssistantSubmit(
```
mocca102, line 334: Optional. But following the pattern you have I believe we can do something like this for the types as well

`trackAssistantSubmit: Tracker['trackAgentSubmit'];`

mocca102 (approved):
Left some totally optional comments. Thank you for working on this. It is looking great!

src/modules/agent.js
evanyan13: Yeah I think that's why I didn't include much description 😓 
And I think to maintain the structure (for search, browse modules etc), I will js leave the description as such for now

esezen (changes requested):
@evanyan13 the changes look good but some tests are failing both in CI and locally. Do you mind taking a look why?

evanyan13:
> @evanyan13 the changes look good but some tests are failing both in CI and locally. Do you mind taking a look why?

Hi @esezen  Thank you for raising the issues!
I have identified the mistake that is causing the failure:
* The `behavioral_action` endpoints have not been updated to receive `agent` so I have reverted to use `assistant` for now. (eg. `assistant_submit`)
* Do we know if the team has plan to update these endpoints to use `agent`?

Also, I'm facing a problem on the [test buildsI](https://github.com/Constructor-io/constructorio-client-javascript/actions/runs/16484014249/job/46606818916?pr=386). Here are my observations:
* All of the tests pass on my local, but some of them seems to fail in the github workflow
* Different number of test cases fail each time (~1-3)?
* I found that `trackAssistantSubmit` and `trackAgentSubmit` are using the same method, but `trackAgentSubmit` fails while `trackAssistantSubmit` passes
So I think this is likely due to a Timeout error rather than an implementation error.
Do you think this is something we should follow up/dig deeper about?

esezen:
@evanyan13 
> Do we know if the team has plan to update these endpoints to use agent?
Can you please check with ASA team? It's not a blocker for this PR and I will merge this PR today but let's figure out the next steps there.

> Different number of test cases fail each time (~1-3)?
We have some flaky tests we need to address but not related to your PR, all good!

esezen (approved):
LGTM!

PR #397, opened 2025-08-19
[CDX-23] Fix errors thrown out by test:types

esezen:


src/types/request-queue.d.ts
```diff
@@ -0,0 +1,29 @@
+import { EventEmitter } from './events';
+
+export default class RequestQueue {
```
esezen, line 3: @Mudaafi do you think we should type this in detail or is it enough for purposes of this story?

jjl014: My 2cents - I think this is probably fine as we probably won't expect folks to make changes to this often, and consumers of the library wouldn't be expected to use it directly. It'd be helpful for future devs that work within the repo, but we can always add more details in the future if needed.

The decision is ultimately up to you guys though. 👍

jjl014 (approved):
The changes look good to me! 🚀

PR #398, opened 2025-08-25
[REM-1993] Add sl tracking in recs result view

lws803:
Updates beacon to include sponsored listings tracking (campaign id, owner) for recommendations view events.

Continuation from https://github.com/Constructor-io/constructorio-client-javascript/pull/389

### DIFF
--- a/src/modules/tracker.js
+++ b/src/modules/tracker.js
@@ -1232,6 +1232,8 @@ class Tracker {
    * @param {string[]|string|number} [parameters.seedItemIds] - Item ID(s) to be used as seed
    * @param {object} [networkParameters] - Parameters relevant to the network request
    * @param {number} [networkParameters.timeout] - Request timeout (in milliseconds)
+   * @param {string} [parameters.slCampaignId] - Pass campaign id of sponsored listing
+   * @param {string} [parameters.slCampaignOwner] - Pass campaign owner of sponsored listing
    * @returns {(true|Error)}
    * @description User viewed a set of recommendations
    * @example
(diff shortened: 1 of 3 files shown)


esezen (dismissed):

esezen (changes requested):
@lws803 the test is failing with "Invalid parameters" error. You can re-create by running the test you modified locally

src/modules/tracker.js
```diff
@@ -1268,6 +1272,8 @@ class Tracker {
         analyticsTags,
         seedItemIds,
         resultCount = result_count || items?.length || 0,
+        slCampaignId,
+        slCampaignOwner,
```
jjl014, line 1276: I'm not sure if we need this here 🤔 

I believe we pass `slCampaignId` and `slCampaignOwner` via the `items` array. That seems to be how home24 has it implemented for Search/Browse based on what I see in their beacon. cc @esezen 

Screenshot 2025-08-27 at 5 10 37 PM

jjl014 (commented):
I'm not sure if we need these changes for the recs result view event.

I think the change would be on the behavioral endpoint to make sure `slCampaignId` or `slCampaignOwner` is supported on the individual item level in the `items` array we send back with the view event.

src/modules/tracker.js
lws803: Hey @jjl014 , just verified this. Thanks for clarifying! Oh yeah it seems that's already supported out of the box for all items schemas. No changes required on either behavioral endpoint or the beacons here 👍.

Will close this PR since it's no longer required.

PR #449, opened 2026-05-01
[NO-REF] - introduce pia http module

vladconstructor:


vladconstructor:
@claude

src/modules/pia.js
```diff
@@ -0,0 +1,185 @@
...
+      const controller = new AbortController();
+
+      signal = controller && controller.signal;
+
+      helpers.applyNetworkTimeout(this.options, networkParameters, controller);
+    }
+
+    if (!question || typeof question !== 'string') {
+      return Promise.reject(new Error('question is a required parameter of type string'));
+    }
```
esezen: Can we move this up to come before AbortController so that we never even instantiate the AbortController if the question is not valid?

src/modules/pia.js
```diff
@@ -0,0 +1,185 @@
...
+    return fetch(requestUrl, { signal })
+      .then(helpers.convertResponseToJson)
+      .then((json) => {
+        if (json.questions) {
+          this.eventDispatcher.queue('pia.getSuggestedQuestions.completed', json);
+
+          return json;
+        }
+
+        throw new Error('getSuggestedQuestions response data is malformed');
```
esezen, line 125: Can we add a test to check for this case?

src/modules/pia.js
```diff
@@ -0,0 +1,185 @@
...
+    queryParams.us = segments;
+  }
+
+  // Pull user id from options and ensure string
+  if (userId) {
+    queryParams.ui = String(userId);
+  }
+
+  if (parameters) {
+    const { threadId, variationId, numResults } = parameters;
```
esezen, line 42: Can we update the tests to check for threadId?

src/modules/pia.js
```diff
@@ -0,0 +1,185 @@
...
+
+    if (threadId) {
+      queryParams.thread_id = threadId;
+    }
+
+    if (variationId) {
+      queryParams.variation_id = variationId;
+    }
+
+    if (!helpers.isNil(numResults)) {
```
esezen, line 52: Same as above, can we test this in one of the tests?

src/types/pia.d.ts
```diff
@@ -0,0 +1,60 @@
...
+  threadId?: string;
+  variationId?: string;
+}
+
+export interface PiaSuggestedQuestionsResponse {
+  questions: Array<PiaQuestion>;
+}
+
+export interface PiaAnswerItemResults {
+  request?: Record<string, any>;
```
esezen: Is this actually very loose or do we know the shape of this? If we do, can we type it strictly? I know we have `Record<string, any>` in other parts of the library but we are trying to move away from that

esezen (dismissed):
Thanks for working on this, I left some small comments.

src/modules/pia.js
```diff
@@ -0,0 +1,185 @@
...
+    apiKey,
+    clientId,
+    sessionId,
+    segments,
+    userId,
+    version,
+    agentServiceUrl,
+    assistantServiceUrl,
+  } = options;
+  const serviceUrl = agentServiceUrl || assistantServiceUrl;
```
Mudaafi, line 17: Will the serviceUrl always be the same as ASA? If they're different, maybe we should surface a separate top-level prop

src/types/pia.d.ts
```diff
@@ -0,0 +1,65 @@
...
+  questions: Array<PiaQuestion>;
+}
+
+export interface PiaAnswerItemResults {
+  request?: {
+    item_id: string;
+    variation_id?: string;
+    num_results?: number;
+    thread_id?: string;
+  };
```
Mudaafi, line 34: hmm, there are other fields in the request object (kind of like SABR), but there's no good type to extend it from. We can leave this as is for now.

spec/src/modules/pia.js
```diff
@@ -0,0 +1,347 @@
...
+        expect(requestedUrlParams).to.have.property('key');
+        expect(requestedUrlParams).to.have.property('i');
+        expect(requestedUrlParams).to.have.property('s');
+        expect(requestedUrlParams).to.have.property('c').to.equal(clientVersion);
+        expect(requestedUrlParams).to.have.property('_dt');
+        expect(requestedUrlParams).to.have.property('item_id').to.equal(validItemId);
+      });
+    });
+
+    it('Should return a result provided a valid apiKey, itemId and variationId', () => {
```
Mudaafi, line 75: it seems like variationId doesn't need to be valid to return a result. Is this expected?

Mudaafi (approved):
lgtm! I had some questions, but if they're expected, we should be good.

src/modules/pia.js
vladconstructor: I think they should always be the same. There are existing customers who are already integrated with PIA via API. I do not expect it to be changed any time soon.

spec/src/modules/pia.js
vladconstructor: Yes, this is expected. The test is verifying that the parameter gets passed through to the API correctly
  (checking requestedUrlParams has variation_id), not that the API validates it. The API simply accepts whatever variationId is sent since it is a filtering/context parameter

PR #455, opened 2026-05-28
[CDX-468] Add modern ESM build

esezen:


package.json
```diff
@@ -17,15 +18,19 @@
...
     "test:all:parallel": "npm run test:parallel && npm run test:bundled:parallel",
     "precoverage": "rm -rf ./coverage && rm -rf ./.nyc_output",
     "coverage": "nyc --all --reporter=html npm run test:src",
     "postcoverage": "open coverage/index.html && rm -rf test",
     "docs": "jsdoc --configure ./.jsdoc.json ./README.md --recurse ./src --destination ./docs",
     "compile": "rm -rf ./lib/* && babel src/ -d lib/ --copy-files && rm -rf ./lib/types/tests",
+    "postcompile": "node ./scripts/build-esm.js",
     "prepublish": "npm run compile",
     "bundle": "rm -rf ./dist/* && npm run compile && node bundle.js",
+    "postbundle": "node ./scripts/bundle-esm.js",
```
Mudaafi: Why `postbundle` and not:
```suggestion
"bundle": "rm -rf ./dist/* && npm run compile && node bundle.js && node ./scripts/bundle-esm.js",
```

Not that the hook doesn't work, it does, but why use the extra brain power?

scripts/bundle-esm.js
```diff
@@ -0,0 +1,30 @@
+const packageJSON = require('../package.json');
```
Mudaafi, line 1: Should this live in `script/s` or should it be top-level like the other `bundle.js` file? Or maybe we should move that one down 🤔

Mudaafi (commented):
@esezen A couple of considerations before we merge it in below. Should we also update the readme with instructions on how to utilize the served bundles?

scripts/bundle-esm.js
esezen: Good catch, I think it makes sense to move the other one down [26d402c](https://github.com/Constructor-io/constructorio-client-javascript/pull/455/commits/26d402cbaf2cdf36a88255e574bd7aaad9afddfb)

Mudaafi (approved):
lgtm, let's merge it on Monday

PR #458, opened 2026-06-03
[AT-82]: add new quiz_asa_results_message type

dmitrycnstrc:
Added new quiz_asa_results_message type for quizzes


src/types/quizzes.d.ts
```diff
@@ -113,6 +113,7 @@ export interface QuizResultsResponse extends Record<string, any> {
     has_attribute: boolean;
     is_matched: boolean;
   }>;
+  quiz_asa_results_message?: Nullable<string>;
```
esezen: Why do we need Nullable here? Would it ever be null?

dmitrycnstrc: I've updated the quiz-service to make it optional instead. It's only included in the response when ASA returns at least one product result along with a message, so there's no scenario where it would be null. If there's no message, the field is simply omitted from the response

esezen (approved):

PR #459, opened 2026-06-05
[CDX-201] Tracking other related parameters

Alexey-Pavlov:
This pull request introduces support for using global window variables as fallback values for `userId`, `testCells`, and `userSegments` in the main query modules (`search`, `browse`, `autocomplete`, and `recommendations`) when the new `trackWindowParameters` option is enabled. It also ensures that the `tracker` module does not use these window globals, and adds comprehensive tests to validate this behavior. Additionally, a utility function for applying these window-based parameter getters is introduced and thoroughly tested.

### DIFF
--- a/spec/src/modules/search.js
+++ b/spec/src/modules/search.js
@@ -150,6 +150,88 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
       });
     });
 
+    it('Should include window global userId when trackWindowParameters is true and options.userId is absent', (done) => {
+      window.cnstrc = { userId: 'window-user-id' };
+      const { search } = new ConstructorIO({
+        apiKey: testApiKey,
+        trackWindowParameters: true,
+        fetch: fetchSpy,
+      });
+
+      search.getSearchResults(query, { section }).then(() => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+        expect(requestedUrlParams).to.have.property('ui').to.equal('window-user-id');
+        delete window.cnstrc;
+        done();
+      });
+    });
+
+    it('Should include window global testCells when trackWindowParameters is true and options.testCells is absent', (done) => {
+      window.cnstrc = { testCells: { experiment: 'variation_a' } };
+      const { search } = new ConstructorIO({
+        apiKey: testApiKey,
+        trackWindowParameters: true,
+        fetch: fetchSpy,
+      });
+
+      search.getSearchResults(query, { section }).then(() => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+        expect(requestedUrlParams).to.have.property('ef-experiment').to.equal('variation_a');
+        delete window.cnstrc;
+        done();
+      });
+    });
+
+    it('Should include window global userSegments when trackWindowParameters is true and options.segments is absent', (done) => {
+      window.cnstrc = { userSegments: ['vip', 'beta'] };
+      const { search } = new ConstructorIO({
+        apiKey: testApiKey,
+        trackWindowParameters: true,
+        fetch: fetchSpy,
+      });
+
+      search.getSearchResults(query, { section }).then(() => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+        expect(requestedUrlParams).to.have.property('us').to.deep.equal(['vip', 'beta']);
+        delete window.cnstrc;
+        done();
+      });
+    });
+
+    it('Should prefer options.userId over window global when trackWindowParameters is true', (done) => {
+      window.cnstrc = { userId: 'window-user-id' };
+      const { search } = new ConstructorIO({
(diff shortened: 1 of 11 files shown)


src/constructorio.js
```diff
@@ -139,14 +141,21 @@ class ConstructorIO {
       beaconMode: (beaconMode === false) ? false : true, // Defaults to 'true',
       networkParameters: networkParameters || {},
       humanityCheckLocation: humanityCheckLocation || 'session',
+      trackWindowParameters: trackWindowParameters || false,
     };
 
+    this.trackerOptions = { ...this.options };
```
esezen: The spread `{ ...this.options }` copies plain values before `applyWindowParameterGetters` installs the getters three lines down. Why do we want to exclude tracker here?

src/constructorio.js
```diff
@@ -169,32 +178,34 @@ class ConstructorIO {
   setClientOptions(options) {
     if (Object.keys(options).length) {
       const { apiKey, segments, testCells, sessionId, userId, sendTrackingEvents } = options;
+      const setOption = (key, value) => {
```
esezen: This is dependant on the previous comment but because there are now two independent options objects, every update has to write to both. Any future property added to a `setClientOptions` path must remember to update `trackerOptions` too, or the two drift apart silently.

src/utils/helpers.js
```diff
@@ -390,6 +390,59 @@ const utils = {
     return filtered;
   },
 
+  applyWindowParameterGetters: (options) => {
```
esezen, line 393: This helper is exported and its getters reference `window` directly. It's only guarded by the `canUseDOM()` check at the call site in `constructorio.js`. If it's ever called independently in a DOM-less context (or a test forgets JSDOM), reading `options.userId/testCells/segments` throws `ReferenceError`. What do you think about adding `canUseDOM()` no-op guard inside the helper so it's safe on its own.

src/utils/helpers.js
```diff
@@ -390,6 +390,59 @@ const utils = {
...
+    const backing = {
+      userId: options.userId || undefined,
+      testCells: options.testCells || undefined,
+      segments: options.segments || undefined,
+    };
+
+    Object.defineProperty(options, 'userId', {
```
esezen: There is a good amount of repeating code with these three. What do you think about extracting some of that logic into a helper?

src/constructorio.js
```diff
@@ -139,14 +141,21 @@ class ConstructorIO {
       beaconMode: (beaconMode === false) ? false : true, // Defaults to 'true',
       networkParameters: networkParameters || {},
       humanityCheckLocation: humanityCheckLocation || 'session',
+      trackWindowParameters: trackWindowParameters || false,
```
esezen: `trackWindowParameters || false` treats any truthy non-boolean (e.g. the string `'false'`) as enabled. And the gate at line 149 reads the raw constructor arg, not the normalized value — so even after normalizing `this.options`, the feature can still switch on from a bad input. Can we do `trackWindowParameters === true`?

spec/src/modules/search.js
```diff
@@ -150,6 +150,88 @@ describe(`ConstructorIO - Search${bundledDescriptionSuffix}`, () => {
...
+      const { search } = new ConstructorIO({
+        apiKey: testApiKey,
+        trackWindowParameters: true,
+        fetch: fetchSpy,
+      });
+
+      search.getSearchResults(query, { section }).then(() => {
+        const requestedUrlParams = helpers.extractUrlParamsFromFetch(fetchSpy);
+        expect(requestedUrlParams).to.have.property('ui').to.equal('window-user-id');
+        delete window.cnstrc;
```
esezen: Several tests set `window.cnstrc` and clean it up with `delete window.cnstrc` _inside_ the `.then()` callback. If an assertion throws first (or the promise rejects), the global leaks into subsequent tests. Can we move cleanup to `afterEach`?

esezen (changes requested):
**1. Does the tracker need window params too?**

Right now `trackerOptions` is snapshotted _before_ `applyWindowParameterGetters` runs (`src/constructorio.js:147`), so the **tracker module is deliberately excluded** from window fallbacks — search/browse/autocomplete/recommendations get them, tracking events don't. Why are we doing this? I thought we wanted to include tracker in this implementation as well.

**2. Renaming `trackWindowParameters`**

The `track` prefix collides with the SDK's tracker vocabulary. What do you think about renaming it? Some options I thought about:

- `useWindowParameters`
- `useDefaultWindowParameters`
- `fallbackToWindowParameters`

Alexey-Pavlov:
Hey @esezen! Thanks for the thoughtful review!
About the tracker module - It all makes sense, I think I just missed it, maybe I was hurrying to do it before my OOO. Thanks for catching it! 
I renamed **trackWindowParameters** to **useWindowParameters** (it's more concise than **fallbackToWindowParameters** and avoids the track* collision), and made other necessary fixes and adjustments. 
Could you please take a look when you get a chance?

esezen (approved):
Thanks for the changes!

PR #463, opened 2026-06-10
Cdx 429 add result offset parameter support to js sdk tracker

Sher-Bakhodirov:
## Summary

Adds support for the `resultOffset` parameter across the tracker functions

## Changes

Added the `resultOffset` parameter to the following tracker functions:
- `trackSearchResultsLoaded`
- `trackBrowseResultsLoaded`
- `trackBrowseResultClick`
- `trackRecommendationView`
- `trackRecommendationClick`
- `trackGenericResultClick`

## Intentionally Excluded

`resultOffset` was **not** added to the following functions:
- `trackSearchResultClick` — uses the legacy backend endpoint, which does not support `result_offset`.

## Test Cleanup

The tests previously used the pattern `Object.assign(requiredParameters, optionalParameters)`, which mutates the original `requiredParameters` object by merging the optional parameters into it. This caused optional parameters to leak into test cases that were only meant to use the required parameters. Updated these spots to avoid mutating the shared object.


src/modules/tracker.js
```diff
@@ -2051,6 +2085,7 @@ class Tracker {
    * @param {string} [parameters.itemName] - Product item name
    * @param {string} [parameters.variationId] - Product item variation unique identifier
    * @param {string} [parameters.section="Products"] - Index section
+   * @param {number} [parameters.resultOffset] - Current offset of results, used on scrolling sites. Cannot be used with `resultPage`
```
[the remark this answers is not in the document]
esezen: @Sher-Bakhodirov I think this makes sense. Wdyt?

spec/src/modules/tracker.js
```diff
@@ -15611,7 +16260,7 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
 
       expect(
         tracker.trackMediaImpressionView(
-          Object.assign(requiredParameters, optionalParameters),
```
esezen, line 15614: Why are we changing these?

src/modules/tracker.js
```diff
@@ -676,6 +676,7 @@ class Tracker {
    * @param {object[]} parameters.items - List of product item unique identifiers in search results listing
```
esezen, line 676: Not related to this line but can we please add offset to `trackSearchResultClickV2` and `trackSearchResultClick` as well?

esezen (changes requested):
PR is looking good overall. I left some comments

spec/src/modules/tracker.js
Sher-Bakhodirov: `Object.assign(requiredParameters, optionalParameters)` mutates `requiredParameters` in place, so optional params leaked into tests that should only use the required ones. The change just avoids mutating the shared object. I can revert if you'd rather keep it out of this PR

src/modules/tracker.js
Sher-Bakhodirov: For `trackSearchResultClick`, it's using a legacy endpoint that doesn't support `result_offset`. And `trackSearchResultClickV2` uses an endpoint that wasn't listed in the ticket, so I think we'd want to verify it supports `result_offset` before adding it. I can include both if you can confirm

src/modules/tracker.js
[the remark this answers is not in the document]
Sher-Bakhodirov: 911e89fcc50987b606ae767e42f6fcff939cab42

spec/src/modules/tracker.js
esezen: Makes sense. We can keep it

src/modules/tracker.js
esezen: If V1 doesn't support it, we can skip it. Let's add it to V2. Thank you!

Sher-Bakhodirov: Sure thing
e299c754b00c7e822c6ec7b29fcf71a737e1d715

esezen (approved):
Thanks for making the changes!

PR #469, opened 2026-06-18
[NO-REF] - add pia recommendations tracking

vladconstructor:


src/types/tracker.d.ts
```diff
@@ -428,7 +434,10 @@ declare class Tracker {
       question: string;
       answerText: string;
       qnaResultId?: string;
+      items?: ItemTracked[];
+      followUpQuestions?: Array<{ question: string }>;
```
TarekAlQaddy: The RFC mentions that `followUpQuestions` is an Array of objects, each object should have a `value`, could you confirm that this was changed to `question` field instead?

src/types/tracker.d.ts
```diff
@@ -441,6 +450,20 @@ declare class Tracker {
...
+  ): true | Error;
+
+  trackProductInsightsAgentResultClick(
+    parameters: {
+      itemId: string;
+      itemName: string;
+      position?: number;
+      qnaResultId?: string;
+      variationId?: string;
+      threadId?: string;
```
TarekAlQaddy, line 470: There are additional params that are mentioned in the RFC which are `question`, `seedItemId`, `seedItemName` and `seedVariationId`. Could you clarify if they were intentionally omitted?

TarekAlQaddy (commented):
Thanks @vladconstructor for working on this!
I left some comments to clarify some params, could you check them please?

spec/src/modules/tracker.js
```diff
@@ -15773,6 +15778,12 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
         expect(requestParams).to.have.property('section').to.equal(optionalParameters.section);
         expect(bodyParams).to.have.property('variation_id').to.equal(optionalParameters.variationId);
         expect(bodyParams).to.have.property('qna_result_id').to.equal(optionalParameters.qnaResultId);
+        expect(bodyParams).to.have.property('thread_id').to.equal(optionalParameters.threadId);
+        expect(bodyParams).to.have.property('items').to.be.an('array').with.lengthOf(2);
+        expect(bodyParams.items[0]).to.have.property('item_id').to.equal('rec1');
+        expect(bodyParams.items[0]).to.have.property('item_name').to.equal('Rec Product 1');
+        expect(bodyParams).to.have.property('follow_up_questions').to.be.an('array').with.lengthOf(2);
+        expect(bodyParams.follow_up_questions[0]).to.have.property('question').to.equal('What about size?');
```
TarekAlQaddy: The test fails as this should be `value` instead of `question`

TarekAlQaddy (changes requested):

TarekAlQaddy (dismissed):
LGTM!
I ran the tests locally and 2 tests were failing in trackMediaImpressionView and trackMediaImpressionClick which are not related to the changes of this PR

src/modules/tracker.js
```diff
@@ -3745,6 +3787,18 @@ class Tracker {
         qna_result_id: qnaResultId,
       };
 
+      if (items && Array.isArray(items)) {
+        bodyParams.items = items.slice(0, 100).map((item) => helpers.toSnakeCaseKeys(item, false));
```
esezen: Slicing is not required anymore. Can we please remove it?

src/modules/tracker.js
```diff
@@ -3745,6 +3787,18 @@ class Tracker {
...
+      }
+
+      if (followUpQuestions && Array.isArray(followUpQuestions)) {
+        bodyParams.follow_up_questions = followUpQuestions.slice(0, 100).map((q) => helpers.toSnakeCaseKeys(q, false));
```
esezen: Same here

spec/src/modules/tracker.js
```diff
@@ -16259,6 +16272,181 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
+    it('Should throw an error when no parameters are provided', () => {
+      const { tracker } = new ConstructorIO({ apiKey: testApiKey });
+
+      expect(tracker.trackProductInsightsAgentResultClick()).to.be.an('error');
+    });
+
+    it('Should throw an error when invalid parameters are provided', () => {
+      const { tracker } = new ConstructorIO({ apiKey: testApiKey });
+
+      expect(tracker.trackProductInsightsAgentResultClick()).to.be.an('error');
```
esezen: Is this a duplicate of `Should throw an error when no parameters are provided`? The test says invalid but this line doesn't pass anything

src/modules/tracker.js
```diff
@@ -3461,6 +3474,7 @@ class Tracker {
         itemId,
         itemName,
         variationId,
+        threadId,
```
esezen, line 3477: Can we update the tests to check for threadId for all functions where we added `threadId`?

src/types/tracker.d.ts
```diff
@@ -428,7 +434,10 @@ declare class Tracker {
       question: string;
       answerText: string;
       qnaResultId?: string;
+      items?: ItemTracked[];
+      followUpQuestions?: Array<{ value: string }>;
```
esezen: Is there a reason why this is not `Question[]`?

src/modules/tracker.js
```diff
@@ -3817,6 +3873,116 @@ class Tracker {
...
+   * @param {string} [parameters.threadId] - Thread ID for grouping events within a conversation
+   * @param {string} [parameters.section] - The section name for the item Ex. "Products"
+   * @param {object} [networkParameters] - Parameters relevant to the network request
+   * @param {number} [networkParameters.timeout] - Request timeout (in milliseconds)
+   * @returns {(true|Error)}
+   * @description User clicked a recommended product in the PIA widget
+   * @example
+   * constructorio.tracker.trackProductInsightsAgentResultClick(
+   *   {
+   *     'itemId': '10',
```
esezen, line 3928: Copilot:

>The JSDoc example for trackProductInsightsAgentResultClick omits question and seedItemId, but both are documented as required parameters. This example is likely to mislead users into sending incomplete payloads.

esezen (changes requested):
Thanks for the PR!

src/types/tracker.d.ts
vladconstructor: Because structures are different. FollowUpQuestion is { value: string } whilst Question is { question: string }. It did not make a lot of sense to reuse existing Question type within the API as it would require to redefine the description for this exact field. I can create a separate type for follow_up_questions though.

esezen (approved):
LGTM!

PR #472, opened 2026-06-22
[CDX-470] Add `additionalTrackingKeys` option

t3t3c:
## Summary

- Adds `additionalTrackingKeys` option to `ConstructorClientOptions` that accepts an array of API key strings
- When configured, every tracking event queued via `RequestQueue` is duplicated for each additional key, with the `key` swapped in both the URL query string and POST body
- Invalid entries (non-strings, empty strings) are filtered out silently

## Usage

```js
const cio = new ConstructorIO({
  apiKey: 'primary-key',
  sendTrackingEvents: true,
  additionalTrackingKeys: ['secondary-key', 'tertiary-key'],
});

// All tracker methods now send events to all 3 keys
cio.tracker.trackPurchase({ items: [...], revenue: 12.00 });
```

## Changes

- `src/types/index.d.ts` — Added `additionalTrackingKeys?: string[]` to options interface
- `src/constructorio.js` — Pass through the new option
- `src/utils/request-queue.js` — Duplicate queued requests for each additional key
- `spec/src/utils/request-queue.js` — Unit tests (5 cases)
- `spec/src/modules/tracker.js` — Integration test

## Test plan

- [ ] Unit tests: duplication works, empty array, missing option, invalid entries, GET requests
- [ ] Integration test: fetch called with both primary and additional key URLs
- [ ] Existing test suite: no regressions
- [ ] Lint passes

Resolves CDX-470


src/utils/request-queue.js
```diff
@@ -53,6 +53,28 @@ class RequestQueue {
         body,
         networkParameters,
       });
+
+      // Duplicate request for each additional tracking key
+      const additionalKeys = this.options?.additionalTrackingKeys;
+
+      if (Array.isArray(additionalKeys) && additionalKeys.length) {
```
esezen: Not a big deal but can we:

1. Make this array into a set to remove duplicates
2. Check the keys in here against the original key to make sure we don't duplicate events

spec/src/modules/tracker.js
```diff
@@ -17075,4 +16345,39 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
+        callCount += 1;
+
+        if (callCount === 2) {
+          expect(fetchSpy).to.have.been.calledTwice;
+
+          const firstCallUrl = fetchSpy.getCall(0).args[0];
+          const secondCallUrl = fetchSpy.getCall(1).args[0];
+
+          expect(firstCallUrl).to.contain(`key=${testApiKey}`);
+          expect(secondCallUrl).to.contain(`key=${additionalKey}`);
```
esezen, line 17344: Can we also check for `not.contain(testApiKey)` here?

spec/src/modules/tracker.js
```diff
@@ -17075,4 +16345,39 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
       ).to.equal(true);
     });
   });
+
+  describe('additionalTrackingKeys', () => {
```
esezen, line 17322: Can we also add a test for verifying the body of the requests are the same in the duplicate request?

src/constructorio.js
```diff
@@ -59,6 +59,7 @@ class ConstructorIO {
    * @param {object} [parameters.networkParameters] - Parameters relevant to network requests
    * @param {number} [parameters.networkParameters.timeout] - Request timeout (in milliseconds) - may be overridden within individual method calls
    * @param {string} [parameters.humanityCheckLocation='session'] - Storage location for the humanity check flag ('session' for sessionStorage, 'local' for localStorage)
+   * @param {string[]} [parameters.additionalTrackingKeys] - Additional API keys to duplicate tracking events to
```
esezen: I am fine with the current version as well but this might explain this concept a bit further

```suggestion
   * @param {string[]} [parameters.additionalTrackingKeys] - Additional API keys that each receive a duplicate of every tracking event. Events are always sent to the primary `apiKey` and, in addition, to every key in this
  array.```

**Review by esezen — CHANGES_REQUESTED** [review]
Thanks! This is looking great. I left some comments

File: src/utils/request-queue.js (hunk 1 above)
Comment (t3t3c) [reply src/utils/request-queue.js]: Good idea! Added it

File: src/constructorio.js (hunk 4 above)
Comment (t3t3c) [reply src/constructorio.js]: Added your suggestion

File: src/constructorio.js [hunk 5]
```diff
@@ -97,6 +98,7 @@ class ConstructorIO {
       networkParameters,
       humanityCheckLocation,
       useWindowParameters,
+      additionalTrackingKeys,
```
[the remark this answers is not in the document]
evanyan13: Hi @t3t3c, I think this is a good point to consider.
If client options like `apiKey` and `sendTrackingEvents` can be updated via `setClientOptions`, maybe we should consider allowing update of the new param `additionalTrackingKeys` via `setClientOptions` as well.
Feel free to let me know what you think?

src/utils/request-queue.js
```diff
@@ -53,6 +53,30 @@ class RequestQueue {
...
+        const encodedOriginalKey = helpers.encodeURIComponentRFC3986(this.options.apiKey);
+
+        const uniqueKeys = new Set(
+          additionalKeys.filter((key) => key && typeof key === 'string' && key !== this.options.apiKey),
+        );
+
+        uniqueKeys.forEach((additionalKey) => {
+          const encodedKey = helpers.encodeURIComponentRFC3986(additionalKey);
```
evanyan13:
```suggestion
          const encodedAdditionalKey = helpers.encodeURIComponentRFC3986(additionalKey);
```
For your consideration, I lost track of the `additionalKey` const as I was reading this portion so maybe it's better to be more explicit in the naming for this const.

spec/src/utils/request-queue.js
```diff
@@ -318,6 +318,121 @@ describe('ConstructorIO - Utils - Request Queue', function utilsRequestQueue() {
...
+
+        requests.queue('https://ac.cnstrc.com/behavior?action=session_start&key=primary-key&_dt=123', 'POST', { action: 'session_start', key: 'primary-key' });
+
+        const queue = RequestQueue.get();
+        expect(queue).to.be.an('array').length(3);
+        expect(queue[1].url).to.contain('key=valid-key');
+        expect(queue[2].url).to.contain('key=another-valid-key');
+      });
+    });
+
```
evanyan13, line 411: Can we add another test to check for the behavior of removing duplicate keys that are the same as the original key?
Since this is an explicit function we added recently

spec/src/utils/request-queue.js
```diff
@@ -318,6 +318,121 @@ describe('ConstructorIO - Utils - Request Queue', function utilsRequestQueue() {
...
+        requests.queue('https://ac.cnstrc.com/behavior?action=session_start&key=primary-key&_dt=123');
+
+        const queue = RequestQueue.get();
+        expect(queue).to.be.an('array').length(2);
+
+        expect(queue[0].url).to.contain('key=primary-key');
+        expect(queue[1].url).to.contain('key=extra-key-1');
+        expect(queue[1].url).to.contain('action=session_start');
+        expect(queue[1].body.key).to.equal('extra-key-1');
```
evanyan13: [This comment above](https://github.com/Constructor-io/constructorio-client-javascript/pull/472/changes#r3578204288) seems to be valid.
Since `sendEvents` completely ignores the `body` if it's a GET request, even when the `body` exists in the request in the `queue` it does not matter in the end.
```
if (nextInQueue.method === 'GET') {
        request = fetch(nextInQueue.url, { signal });
      }
```
Given that, I think we can safely remove **line 415**

spec/src/modules/tracker.js
```diff
@@ -17318,4 +17318,84 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
+          expect(fetchSpy).to.have.been.calledTwice;
+
+          const firstCallOptions = fetchSpy.getCall(0).args[1];
+          const secondCallOptions = fetchSpy.getCall(1).args[1];
+
+          const firstBody = JSON.parse(firstCallOptions.body);
+          const secondBody = JSON.parse(secondCallOptions.body);
+
+          expect(firstBody.key).to.equal(testApiKey);
+          expect(secondBody.key).to.equal(additionalKey);
```
evanyan13, line 17381: Should we add the check here as well
```
expect(secondCallUrl).to.not.contain(`key=${testApiKey}`);
```

evanyan13 (commented):
Hey @t3t3c 
Thank you so much for making the changes! The PR is looking great in general. 🔥 
Just left some comments for your consideration please

src/constructorio.js
```diff
@@ -182,7 +185,7 @@ class ConstructorIO {
    */
```
evanyan13, line 186:
```suggestion
* @param {string[]} [options.additionalTrackingKeys] - Additional API keys that each receive a duplicate of every tracking event. Events are always sent to the primary `apiKey` and, in addition, to every key in this array.
```
Can we add the documentation for the `setClientOptions` method as well

src/constructorio.js
```diff
@@ -214,6 +217,10 @@ class ConstructorIO {
         const formattedServiceUrl = helpers.addHTTPSToString(normalizedServiceUrl, this.options.allowHttpServiceUrl);
         this.options.serviceUrl = formattedServiceUrl || this.options.serviceUrl;
       }
+
+      if (Array.isArray(additionalTrackingKeys)) {
```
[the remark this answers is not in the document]
evanyan13: I think the comment is valid too
Maybe we should add a test in `spec/src/constructorio.js`?
What do you think?
`it('Should use additionalTrackingKeys updated via setClientOptions for subsequent events',`

evanyan13 (commented):
Hi @t3t3c 
Thanks for making the changes so promptly!
Have left some comment regarding the new update to `setClientOptions`
For your consideration please

evanyan13 (dismissed):
LGTM!
Thanks @t3t3c for the PR and the changes
@jjl014 Could we get your review & approval to push this change please
Thank you!

src/utils/request-queue.js
```diff
@@ -53,6 +53,30 @@ class RequestQueue {
...
+      if (Array.isArray(additionalKeys) && additionalKeys.length) {
+        const encodedOriginalKey = helpers.encodeURIComponentRFC3986(this.options.apiKey);
+
+        const uniqueKeys = new Set(
+          additionalKeys.filter((key) => key && typeof key === 'string' && key !== this.options.apiKey),
+        );
+
+        uniqueKeys.forEach((additionalKey) => {
+          const encodedAdditionalKey = helpers.encodeURIComponentRFC3986(additionalKey);
+          const swappedUrl = url.replace(`key=${encodedOriginalKey}`, `key=${encodedAdditionalKey}`);
```
jjl014, line 65: Any reason why we're doing this validation and uniqueness check every time we queue an event?

It feels like an additional step that could be handled earlier (e.g. during client instantiation in the construtor or when setClientOptions is called)?

spec/src/constructorio.js
```diff
@@ -918,6 +918,22 @@ if (!bundled) {
...
+          sessionId: 1,
+        });
+
+        expect(instance.options.additionalTrackingKeys).to.be.undefined;
+
+        instance.setClientOptions({
+          additionalTrackingKeys: ['extra-key-1', 'extra-key-2'],
+        });
+
+        expect(instance.options).to.have.property('additionalTrackingKeys').to.deep.equal(['extra-key-1', 'extra-key-2']);
```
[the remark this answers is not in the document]
jjl014: Claude raised a good point. We should have equivalent tests for the DOM context.

src/constructorio.js
```diff
@@ -214,6 +218,10 @@ class ConstructorIO {
...
```
jjl014: Not a big deal, but if we wanted to stop sending events to multiple keys, we would have to send `[]` at the moment. Should also support `null` here?

jjl014 (commented):
Looking pretty solid to me. I did leave a couple comments that I'd like to get your thoughts on. Thanks!

src/constructorio.js
evanyan13: Yes I think this is a good point!
We should allow users to stop sending events to additional keys by passing `null` to `additionalTrackingKeys` , instead of forcing them to send an empty array `[]`

src/utils/request-queue.js
evanyan13: Thanks @jjl014 This is a good consideration!
We can definitely handle the sanitization check earlier in the system and save the time to conduct the check every time we queue an event 👍

evanyan13:
Hi @jjl014 
Thank you for the feedback! @t3t3c I have updated the PR based on Jimmy's review.
For your review and approval please.
This is a summary of the changes from my end:
* Moved sanitization logic to higher level (at `ConstructorIO.init` and `setClientOptions`) with `helpers.toValidAdditionalTrackingKeys`
* Support empty array and `null` on `additionalTrackingKeys` param as an action to stop sending events to multiple keys
* **`setClientOptions` re-validates on apiKey change:** updating the primary key re-runs validation so a now-duplicate additional key is dropped.
* **Tests restructured:** normalization/clear/re-validation coverage moved up to the ConstructorIO level. Refactor the assertion with `expectSharedOption` helper (asserts the value propagates across all sub-modules), and the request-queue spec keeps only pre-normalized duplication cases.
cc - @t3t3c

jjl014 (commented):
Thanks for making the updates! The changes are looking good from what I can tell, but I noticed that some of the newly added tests and a variety of tests from other parts of the spec are failing.

I'm guessing the ones failing on master might be expected. I think @Mudaafi mentioned he was working on fixing them. However, I wouldn't expect the tests added from this PR to fail as well. 🤔

spec/src/constructorio.js
```diff
@@ -19,6 +19,18 @@ const clientVersion = 'cio-mocha';
 const bundled = process.env.BUNDLED === 'true';
 const bundledDescriptionSuffix = bundled ? ' - Bundled' : '';
 
+// Assert an option on the client and every sub-module that shares its `options` reference.
+// Uses `deep.equal` to enforce strict structural equality for objects and arrays.
+const expectSharedOption = (instance, option, expected) => {
```
Mudaafi:
```suggestion
const expectSharedOption = (instance, key, expectedValue) => {
```

Mudaafi: Good refactor otherwise

spec/src/modules/tracker.js
```diff
@@ -17318,4 +17318,87 @@ describe(`ConstructorIO - Tracker${bundledDescriptionSuffix}`, () => {
...
+        ...requestQueueOptions,
+        additionalTrackingKeys: [additionalKey],
+      });
+
+      let callCount = 0;
+
+      const checkComplete = () => {
+        callCount += 1;
+
+        if (callCount === 2) {
```
Mudaafi, line 17358: shall we add an additional if-else to ensure that it isn't called > 2 times?

Mudaafi (commented):
Looks good, just a couple of comments

spec/src/modules/tracker.js
t3t3c: Added

Mudaafi (approved):
Thanks for making the changes

jjl014 (approved):
LGTM! 🚀

PR #474, opened 2026-07-09
fix: don't mutate caller's fmtOptions when hiddenFields is passed

CedricConday:
## Problem

When a request is made with **both** `fmtOptions` and `hiddenFields`, the request builders assign the caller's `fmtOptions` object to `queryParams.fmt_options` **by reference** and then write `hidden_fields` onto it:

```js
if (fmtOptions) {
  queryParams.fmt_options = fmtOptions;          // same reference as the caller's object
}
if (hiddenFields) {
  if (queryParams.fmt_options) {
    queryParams.fmt_options.hidden_fields = hiddenFields;   // <-- mutates the caller's object
  }
}
```

So calling e.g. `getAutocompleteResults` **mutates the object the caller passed in**. Reusing the same `fmtOptions` object across calls silently accumulates a `hidden_fields` key.

### Reproduction

```js
const fmtOptions = { fields: ['id', 'name'] };
client.autocomplete.getAutocompleteResults('shoes', { fmtOptions, hiddenFields: ['price', 'stock'] });
console.log(fmtOptions);
// { fields: ['id','name'], hidden_fields: ['price','stock'] }  ← caller's object was mutated
```

## Fix

Shallow-copy `fmtOptions` on assignment so the serialized request is **identical** while the caller's input is left intact:

```js
queryParams.fmt_options = { ...fmtOptions };
```

Applied to the five modules that exhibit the mutation: **autocomplete, search, browse, recommendations, quizzes**. `agent.js` assigns by reference too but never writes to it, so it is unaffected and left unchanged.

## Tests

Adds `spec/src/modules/fmtOptions-immutability.js` — an offline regression spec (stubbed `fetch`, no API key/network) asserting the caller's `fmtOptions` is not mutated across all five modules. Verified it **fails on the current code** and **passes with the fix**. `eslint` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DbUHs855FyBt2J7cG2Qmhh

### DIFF
--- a/spec/src/modules/fmtOptions-immutability.js
+++ b/spec/src/modules/fmtOptions-immutability.js
@@ -7,24 +7,13 @@ const ConstructorIO = require('../../../test/constructorio'); // eslint-disable-
 // Regression guard for the `hidden_fields` side-effect: passing both
 // `fmtOptions` and `hiddenFields` must not mutate the caller's `fmtOptions`.
 describe('ConstructorIO - fmtOptions immutability', () => {
-  let fetchStub;
-
-  beforeEach(() => {
-    fetchStub = sinon.stub(global, 'fetch').resolves({
-      ok: true,
-      status: 200,
-      json: () => Promise.resolve({}),
-    });
-  });
-
-  afterEach(() => {
-    fetchStub.restore();
-  });
-
-  const client = () => new ConstructorIO({
+  // Inject the fetch stub through the client options, as the other module specs
+  // do (e.g. spec/src/modules/search.js), rather than stubbing the global.
+  const client = (fetch) => new ConstructorIO({
     apiKey: 'key_immutability_test',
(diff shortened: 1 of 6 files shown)


CedricConday:
Checking back in — still current against `master`.

`src/modules/autocomplete.js:97` assigns the caller's `fmtOptions` to `queryParams.fmt_options` by reference, then line 103 writes `hidden_fields` onto it, so the object the caller passed in is mutated. The same pattern is in `browse.js`, `quizzes.js`, `recommendations.js` and `search.js`, which is why the fix touches all five.

One thing worth flagging: no CI has ever run on this PR, so the checks list is empty rather than green. I think that needs a maintainer to approve the workflow for a fork. Happy to rebase if that helps kick it off.

PR #475, opened 2026-07-17
[AT-176]: Add questionTopic parameter to PIA question tracking events

dmitrycnstrc:
LINEAR: https://linear.app/constructor/issue/AT-176/js-client-add-question-topic-to-pia-tracking-events


Mudaafi (approved):
tests pass

PR #480, opened 2026-07-29
[CDX-398] Backwards Compatibility Fix for CJS users

Mudaafi:
When we released the ESM bundled version earlier, we unintentionally made it the default export. This resulted in CJS environments breaking when they upgraded to the newer minor version since `require` now returns `{ default: xxx }`.

This patch fix removes the typical ESM-structure in package.json in favor of `exports` which allows us to retain the typical `require` pattern while also supporting the usual `import` pattern for ESM. There are limitations however, namely:

1. **Extension-less deep imports are no longer supported**
   1. require('@constructor-io/constructorio-client-javascript/lib/src/modules/search') → MODULE_NOT_FOUND.
   2. Deep imports must also include the extension, i.e. `.js`

2. **Shared types declaration is ambiguous and might fail for `node16` ESM consumers**
    1. Since the JS uses `module.exports` but the types use `export default`, TypeScript under moduleResolution: "node16" might think .default is required when it isn't.
    2. Might need to split types into `.ts` and `.mts`

3. **Older bundlers that do not support `exports` will not work for ESM environments since package.json no longer has the `module` field**

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,8 +3,17 @@
   "version": "2.90.0",
   "description": "Constructor.io JavaScript client",
   "main": "lib/constructorio.js",
-  "module": "lib/esm/constructorio.js",
   "types": "lib/types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./lib/types/index.d.ts",
+      "import": "./lib/esm/constructorio.mjs",
+      "require": "./lib/constructorio.js",
+      "default": "./lib/constructorio.js"
+    },
+    "./lib/*": "./lib/*",
+    "./package.json": "./package.json"
+  },
   "scripts": {
     "clean": "sudo rm -rf node_modules package-lock.json",
     "version": "node ./src/generateVersion.js && npm run docs && git add ./docs/* && npm run bundle && git add -A ./dist && git add ./src/version.js",
--- a/scripts/build-esm.js
+++ b/scripts/build-esm.js
@@ -8,5 +8,7 @@ require('esbuild').build({
   define: {
     global: 'window',
   },
-  outfile: './lib/esm/constructorio.js',
+  // .mjs so Node classifies this as ESM without `"type": "module"`, which would
+  // reclassify every .js in the package and break CJS consumers.
+  outfile: './lib/esm/constructorio.mjs',
 }).catch(() => process.exit(1));


package.json
```diff
@@ -3,8 +3,17 @@
   "version": "2.90.0",
   "description": "Constructor.io JavaScript client",
   "main": "lib/constructorio.js",
-  "module": "lib/esm/constructorio.js",
   "types": "lib/types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./lib/types/index.d.ts",
+      "import": "./lib/esm/constructorio.mjs",
```
Alexey-Pavlov, line 10: There's no `node` condition here. So `import` in Node, and also webpack/Next.js server builds now get the browser bundle (`platform: 'browser'`, `global` replaced with `window`). Before this PR they got the CJS build, because `module` was only read by bundlers.
https://github.com/Constructor-io/constructorio-client-javascript/blob/cdx-498-client-js-fix-require-resolution-broken-by-the-2900-esm/src/constructorio.js#L108
`typeof window !== "undefined" && window.CLIENT_VERSION`



```suggestion
      "node": "./lib/constructorio.js",
      "import": "./lib/esm/constructorio.mjs",
```

package.json
```diff
@@ -3,8 +3,17 @@
...
+      "require": "./lib/constructorio.js",
+      "default": "./lib/constructorio.js"
+    },
+    "./lib/*": "./lib/*",
```
Alexey-Pavlov, line 14: You already listed this as limitation number 1, but I think we can just fix it

```suggestion
    "./lib/*": "./lib/*.js",
    "./lib/*.js": "./lib/*.js",
    "./lib/*.mjs": "./lib/*.mjs",
    "./lib/*.d.ts": "./lib/*.d.ts",
    "./lib/*.map": "./lib/*.map",
```

package.json
```diff
@@ -3,8 +3,17 @@
   "version": "2.90.0",
   "description": "Constructor.io JavaScript client",
   "main": "lib/constructorio.js",
-  "module": "lib/esm/constructorio.js",
```
Alexey-Pavlov, line 6: Modern bundlers ignore the module when exports match, so keeping it changes nothing for them. It only helps old tooling, which just reads the fields and now falls back to CJS

Can we just put it back as 
`"module": "lib/esm/constructorio.mjs"`? 
Then limitation #3 is gone then

package.json
```diff
@@ -3,8 +3,17 @@
...
   "types": "lib/types/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./lib/types/index.d.ts",
```
Alexey-Pavlov, line 9: Maybe, adding something like`npx @arethetypeswrong/cli --pack .` to CI would have caught the 2.90.0 bug too 🤔

https://www.npmjs.com/package/@arethetypeswrong/cli

Alexey-Pavlov (changes requested):
@Mudaafi Thanks for working on this! Left a few comments, please take a look when you get a chance 
Analyzing this PR is quite challenging because there are a lot of different cases, but I think we can fix limitations 1 and 3

package.json
Mudaafi: huh. we'll have to consider other extensions as well like `.css` but I'll dig into it

package.json
Mudaafi: interesting. I would've thought that Node environments would use the Node SDK, but no harm keeping it consistent post-change unless it requires something weird. Good catch

package.json
Mudaafi: I considered that actually. I discarded it though, but I'll need to dig up the reason

PR #482, opened 2026-08-04
add thread id to tracking for asa

v-i-s-h-n-u-ps:


Alexey-Pavlov (dismissed):
LGTM! Maybe it makes sense to add
`@param {string} [parameters.threadId] - Thread ID for grouping events within a conversation`
 to the 6 deprecated trackAssistant* wrappers, but it's a nitpick
For this PR, we would need an additional approval from one of the shepherds

esezen (approved):

PR #484, opened 2026-08-20
[NO-REF] - update PIA module with guard/prefilter/fmtOptions

vladconstructor:


src/modules/pia.js
```diff
@@ -33,13 +34,29 @@ function createPiaUrl(itemId, parameters, options, questionPath) {
     queryParams.us = segments;
   }
 
+  // Pull test cells from options
+  if (testCells) {
```
vladconstructor: I am not sure if this part is needed as PIA API might not support it, but it seems to be part of the legacy code included into every API module

v-i-s-h-n-u-ps (approved):

src/types/pia.d.ts
```diff
@@ -10,15 +12,22 @@ export interface PiaQuestion {
   value: string;
 }
 
-export interface PiaSuggestedQuestionsParameters {
+export interface PiaBaseParameters {
   threadId?: string;
   variationId?: string;
+  features?: Record<string, boolean>;
```
TarekAlQaddy: We typically don't expose `features` and `featureVariants` and we use `qsParam` instead whenever there's a need, since they're kind of internal, mostly used for debugging and not standardized params that we need to maintain moving forward

TarekAlQaddy (changes requested):

TarekAlQaddy (approved):
LGTM

Mudaafi (approved):

PR #485, opened 2026-08-25
Add source parameter to ASA submit tracking

dmitrycnstrc:
Adds an optional source parameter to trackAgentSubmit / trackAssistantSubmit, forwarded to the behavioral-actions API as source in the request body.


HHHindawy (approved):

src/modules/tracker.js
HHHindawy, line 3158: Could we please also document the source param here?

HHHindawy (changes requested):

HHHindawy (approved):

PR #487, opened 2026-09-09
[AT-230]: Add follow_up_refinement agent event type

dmitrycnstrc:
The agent now emits a `follow_up_refinement` SSE event (a narrowing question with
selectable options). The Agent module only subscribes to event types listed in
`Agent.EventTypes`, so the event was silently dropped from the stream.


niizom (approved):
Looks good to me!

dmitrycnstrc:
> Looks good to me!

@niizom 
Can we merge & deploy? Or who should I ping about it?

PR #489, opened 2026-09-15
[AT-217]: support ai_agent_button_click for ASA

dmitrycnstrc:
Adds `tracker.trackAgentButtonClick` to send the new `ai_agent_button_click` behavioral event (Behavioral API v2). It tracks clicks on CTAs that open or start an agent experience, so we can measure ASA entry points and click-to-submit conversion.

```js
tracker.trackAgentButtonClick({
  mode: 'chat',              // required
  domain: 'explorer',        // required
  positionOnPage: 'search_bar',
  pageType: 'pdp',           // home | plp | pdp | collection | email_campaign | cart
  instanceId: 1,             // 1-based
});

### CODE REVIEW

File: src/modules/tracker.js [hunk 1]
```diff
@@ -3140,6 +3140,95 @@ class Tracker {
...
+
+      if (positionOnPage) {
+        bodyParams.position_on_page = positionOnPage;
+      }
+
+      if (pageType) {
+        bodyParams.page_type = pageType;
+      }
+
+      if (instanceId) {
```
HHHindawy: I know the `instanceId` is 1-based, but nothing stops you from passing a `0`.  Should we throw an error in that case, or are we fine with silently ignoring it?

HHHindawy (approved):

